PR #24, opened 2015-09-05
finish object cache and webcal implementation

georgehrke:
object-cache related stuff
~~can't merge, until we have a workaround for owncloud/core#18224~~ (fixed with https://github.com/owncloud/core/pull/18839)


raghunayyar:
Works fine :+1:

PR #28, opened 2015-09-06
Adds spinner to the sidebar before the calendars load.

raghunayyar:
A fix for https://github.com/owncloud/calendar-rework/issues/4

We always had a loader in place, just that we never called it in our template.
Have a look @georgehrke

### DIFF
--- a/css/calendar.css
+++ b/css/calendar.css
@@ -26,7 +26,5 @@
   margin-left: 250px;
 }
 #app-content.loading {
-  background-image: url("../img/loading.gif");
-  background-position: center;
-  background-repeat: no-repeat;
+  background-size: 25px;
 }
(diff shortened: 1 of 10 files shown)


georgehrke:
Two things:
- [ ] I think the spinner is far to big. should not be bigger than an list-item imho
- [ ] there is some time span between the server answers the request and the list is rendered. The list should be rendered first thing once the request was answered, please check.

georgehrke:
- [ ] There is a spinner above fullcalendar, please remove it.

georgehrke:
I changed this pr, looks good from my pov now.
please review and feel free to merge afterwords.

raghunayyar:
Works for me, I also see this as faster than it was previously. :+1: x 2.

PR #36, opened 2015-09-17
Updates Calendar color.

raghunayyar:
- [x] Update calendar colour.
- [x] Broadcast the change on the fullcalendar. cc @georgehrke


georgehrke:
your changes look good, please review mine

georgehrke:
Some dependency is missing, that's why the test fail. please check

raghunayyar:
Travis is giving me : Warning: Cannot find module 'jasmine-core' Use --force to continue.

I don't think this PR has anything to do with the missing deps, I am getting this in. We can get the deps figured out in another PR.

PR #37, opened 2015-09-22
Exception on 4byte utf-8 characters in local db

### LINKED ISSUE #20: Throw exception when adding objects containing emojis to database backend
emojis and other 4 byte unicode characters

codeling:
Should fix #20.
Not tested except for unit tests yet (@georgehrke what API
Also need to elaborate create/update tests for non-error / other error cases.

### DIFF
--- a/backend/local/object.php
+++ b/backend/local/object.php
@@ -28,6 +28,7 @@ namespace OCA\Calendar\Backend\Local;
 use OCA\Calendar\Backend as BackendUtils;
 use OCA\Calendar\Db\ObjectFactory;
 use OCA\Calendar\Db\Permissions;
+use OCA\Calendar\Utility\Utility;
 use OCA\Calendar\Utility\ObjectUtility;
 use OCA\Calendar\Backend\Exception as BackendException;
 use OCA\Calendar\CorruptDataException;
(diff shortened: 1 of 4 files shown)


raghunayyar:
@codeling is this is WIP or a complete PR?

codeling:
Should be complete but have not had a chance to test except for unit tests

DeepDiver1975:
Instead of throwing an exception you could also encode the strings before writing them to the db.

codeling:
@DeepDiver1975 is that how it's done in core? How would it be encoded best?

codeling:
@DeepDiver1975 any thoughts?
@raghunayyar @georgehrke sorry extremely busy otherwise at the moment, didn't have time yet to look more into testing this

DeepDiver1975:
> is that how it's done in core?

As of today core has no mechanism to take care about this.



> How could it be encoded best?

Something we need to look into - thb

georgehrke:
@DeepDiver1975 IIRC you once suggested url encoding.
Yet another question is how to handle existing objects.
Encode all on upgrade?

DeepDiver1975:
> Encode all on upgrade?

seems so

georgehrke:
obsolete here, still very relevant in https://github.com/owncloud/core/issues/20047

>> Encode all on upgrade?

> seems so

Moving to the dav app seems a good opportunity to do so ;)

PR #38, opened 2015-10-18
Set max-version to 9.0

georgehrke:
make it possible to enable the calendar app on ownCloud 9.

please review @raghunayyar


raghunayyar:
Lets go, lets go!

PR #43, opened 2015-10-28
Removes appframework dependancy from 3rdparty.

raghunayyar:
Removes Appframework dependency from 3rdparty.

@georgehrke lets get this one in. :D

PR #44, opened 2015-10-29
Missed one slidetoggle case.

raghunayyar:
This was missing in the #43 PR. @georgehrke please get this one in as well.


georgehrke:
:+1:

PR #60, opened 2015-11-25
added more and optional colors to color picker

EtlamGit:
Based on the discussion in [issue #460 of the original calendar app](https://github.com/owncloud/calendar/issues/460) more colors are defined.
Basis is a ColorWheel with 12 segments + Gray. Starting from that pure colors, darker and less saturated variants are generated in steps of 10%.

As not everybody may like that much colors, most of them are commented out. But as they are provided with the source, it is easy to enable them just by removing the comment.
colorpicker12_20b_rework
There is still a problem in the existing layout between 'new calendar' and 'modify calendar', as the modify has a scroll bar and looks scrambled. When somebody fixes that, it could be considered to force a colorpicker layout with 13 colors in one row. Any hints how to do that would be appreciated.
Comments are welcome!

### DIFF
--- a/css/calendarlist.css
+++ b/css/calendarlist.css
@@ -207,7 +207,7 @@
 /* ColorPicker overrides. */
 .colorpicker {
   display: block;
-  height: 30px;
+  height: 18px;
 }
 .colorpicker .colorpicker-list {
   display: block;
@@ -216,7 +216,7 @@
 }
 .colorpicker .colorpicker-list li {
   height: 100%;
-  width: 29px !important;
+  width: 18px !important;
   float: left;
 }
 .colorpicker .colorpicker-list li.selected {
--- a/js/public/app.js
+++ b/js/public/app.js
@@ -994,7 +994,7 @@ app.controller('SettingsController', ['$scope', '$rootScope', 'Restangular', 'Ca
 app.controller('SubscriptionController', ['$scope', '$rootScope', '$window', 'SubscriptionModel', 'CalendarModel', 'Restangular',
 	function ($scope, $rootScope, $window, SubscriptionModel, CalendarModel, Restangular) {
 		'use strict';
-		
+
 		$scope.subscriptions = SubscriptionModel.getAll();
 		var subscriptionResource = Restangular.all('subscriptions');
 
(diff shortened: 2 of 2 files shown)


georgehrke:
This is a complete overkill from my point of why. 

Additionally editing the app.js is wrong, please see https://github.com/owncloud/calendar-rework#building-javascript

EtlamGit:
It is probably overkill to ship all that colors to every user. But there are some individuals demanding for that much colors. In my personal opinion the colors shown in the image above (and therefore only the not uncommented ones) should be enough for nearly all use cases. The others were intended to allow the power users to easily extend the number of colors while maintaining a well founded selection of the colors based on color science.

From a design standpoint even the colors in the image above are too much. A better solution would be to integrate a color definition tool. In the current calendar app there was a 3rd party app to fulfill that task (but not integrated). I was close to start some experiments to get it running when I noticed that the development moved to calender-rework.
At the end I could imagine to have a very low amount of default colors (e.g. one row) and a color definition button to select a completely user defined color. Maybe only the color definition button is enough. But I have to see it to ground an opinion.

Sorry, I did not know that with node and npm. I will take a look into that.
But it could take some while, as I do not have much spare time...

georgehrke:
We had a color-picker in rework, but it broke due to some change in core and we wanted something working, so we replaced it with the ordinary color picker.
It's still planned to replace it with a proper color picker.

With due respect to your work, but this is a complete overkill and therefore a :-1:.

georgehrke:
@raghunayyar What's your opinion?

raghunayyar:
@EtlamGit I am sorry but I also think I will turn this down, though we definitely need more colour options but this is not the way look at it. As @georgehrke suggested, do you want to work on porting an ng-colorpicker for us? It will completely solve our problem. cc @jancborchardt before closing this PR I would like a word from you too. 

PS: Great to see your PR, welcome to the community.

jancborchardt:
As @georgehrke and @raghunayyar said before, this is too detailed. Here’s what we really need for a color picker:

- about 12 really nice colors
- using the same color palette as for the avatar placeholders: https://github.com/owncloud/core/blob/master/core/js/placeholder.js#L58
- also defining a text color for each of those colors – for some black will be better, for some white will be needed
- for any new calendar which is created, pick the next color which hasn’t been used yet on another calendar by default

jancborchardt:
Also, that color picker should not be hardcoded inside the Calendar but rather be made as its own library so other apps can also potentially use it.

EtlamGit:
I would suggest this color picker: http://labs.abeautifulsite.net/jquery-minicolors/without-bootstrap.html
__It is already included in the current calendar app as 3rd party code__, but not used. As far as I understand the git history, it was also included in owncloud/core some time ago.
 * with bout 2-5 unique colors per user => 12 is the minimum for a family installation
 * with 3 primary colors, 3 secondary colors and 6 in between you get exactly 12 evenly distributed colors from the color wheel 
 * black/white text decision based on luminance is already in the current calendar code

As you search for a developer doing that integration, some words about me: I have a good understanding of html and css. Therefore I did the code changes like they are in this pull request: based on simple css changes. Regarding JavaScript and PHP, I have no experience coding with it (but dozens others like C++ or VHDL). Therefore I assume that I'm not suited to do that task. At least you would get more trouble than solution in that case. Also my spare time is extremely limited, therefore I would not be able to follow your fast development cycle. Sorry, but probably I will not of big help here.

In the future I will continue to write bug reports and create tiny pull requests (that get closed short after; but may start a fruitful discussion...)


Just for the records: I changed the code in my branch to address the issues mentioned here. So in case you could not find someone implementing a nice color picker until release 9.0, you may consider to put this reduced color set as a temporary solution. (Or use any other combination of the defined color rows):
colorpicker12_3_rework

ping: @georgehrke @raghunayyar @jancborchardt I wish you a nice weekend.

jancborchardt:
@EtlamGit as I said above in https://github.com/owncloud/calendar-rework/pull/60#issuecomment-160068121 it’s very important for any color that we use to comply with the existing color palette we use in ownCloud. These colors are too primary, and too many. They do not fit into ownCloud. Please check out the placeholder library I referenced and see which colors it produces. :)

PR #73, opened 2015-12-01
Make calendar app a CalDAV client

georgehrke:
fixes https://github.com/owncloud/calendar-rework/issues/46

### DIFF
--- a/js/app/service/calendarservice.js
+++ b/js/app/service/calendarservice.js
@@ -118,25 +118,7 @@ app.service('CalendarService', ['DavClient', 'Calendar', function(DavClient, Cal
 					continue;
 				}
 
-				var canWrite = false;
-				var acl = body.propStat[0].properties['{' + DavClient.NS_DAV + '}acl'];
-				if (acl) {
-					for (var k=0; k < acl.length; k++) {
-						var href = acl[k].getElementsByTagNameNS('DAV:', 'href');
-						if (href.length === 0) {
-							continue;
-						}
-						href = href[0].textContent;
-						if (href !== _this._currentUserPrincipal) {
-							continue;
-						}
-						var writeNode = acl[k].getElementsByTagNameNS('DAV:', 'write');
-						if (writeNode.length > 0) {
-							canWrite = true;
-						}
-					}
-				}
-				body.propStat[0].properties['canWrite'] = canWrite;
+				_this._getACLFromResponse(body);
 
 				var calendar = new Calendar(body.href, body.propStat[0].properties);
 				calendars.push(calendar);
@@ -165,6 +147,8 @@ app.service('CalendarService', ['DavClient', 'Calendar', function(DavClient, Cal
 				//TODO - something went wrong
 				return;
 			}
+
+			_this._getACLFromResponse(body);
 
 			return new Calendar(body.href, body.propStat[0].properties);
 		});
@@ -293,6 +277,28 @@ app.service('CalendarService', ['DavClient', 'Calendar', function(DavClient, Cal
 				}
 				return cComponents;
 		}
+	};
+
+	this._getACLFromResponse = function(body) {
+		var canWrite = false;
+		var acl = body.propStat[0].properties['{' + DavClient.NS_DAV + '}acl'];
+		if (acl) {
+			for (var k=0; k < acl.length; k++) {
+				var href = acl[k].getElementsByTagNameNS('DAV:', 'href');
+				if (href.length === 0) {
+					continue;
+				}
+				href = href[0].textContent;
+				if (href !== _this._currentUserPrincipal) {
+					continue;
+				}
+				var writeNode = acl[k].getElementsByTagNameNS('DAV:', 'write');
+				if (writeNode.length > 0) {
+					canWrite = true;
+				}
+			}
+		}
+		body.propStat[0].properties['canWrite'] = canWrite;
 	};
 
 	this._isUriAlreadyTaken = function(uri) {
--- a/js/public/app.js
+++ b/js/public/app.js
@@ -1395,25 +1395,7 @@ app.service('CalendarService', ['DavClient', 'Calendar', function(DavClient, Cal
 					continue;
 				}
 
-				var canWrite = false;
-				var acl = body.propStat[0].properties['{' + DavClient.NS_DAV + '}acl'];
-				if (acl) {
-					for (var k=0; k < acl.length; k++) {
-						var href = acl[k].getElementsByTagNameNS('DAV:', 'href');
-						if (href.length === 0) {
-							continue;
-						}
-						href = href[0].textContent;
-						if (href !== _this._currentUserPrincipal) {
-							continue;
-						}
-						var writeNode = acl[k].getElementsByTagNameNS('DAV:', 'write');
-						if (writeNode.length > 0) {
-							canWrite = true;
-						}
-					}
-				}
-				body.propStat[0].properties['canWrite'] = canWrite;
+				_this._getACLFromResponse(body);
 
 				var calendar = new Calendar(body.href, body.propStat[0].properties);
 				calendars.push(calendar);
@@ -1442,6 +1424,8 @@ app.service('CalendarService', ['DavClient', 'Calendar', function(DavClient, Cal
 				//TODO - something went wrong
 				return;
 			}
+
+			_this._getACLFromResponse(body);
 
 			return new Calendar(body.href, body.propStat[0].properties);
 		});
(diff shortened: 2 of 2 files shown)


DeepDiver1975:
Nice! Go go go!

DeepDiver1975:
NICE - that beast is already retrieving calendars from the core dav endpoint!

@georgehrke you rock!

georgehrke:
I have deletion ready and update (prop patch) almost working with my local changes. The EventService should be much easier, because it's without all this xml foo

DeepDiver1975:
> I have deletion ready and update (prop patch) almost working with my local changes. The EventService should be much easier, because it's without all this xml foo

Nice - I just tried to ping you on IRC - I'm planning to invest some time into calendar/dav this week - let me know where I can help out

georgehrke:
> I'm planning to invest some time into calendar/dav this week

That would be great! Would be great if you could take care of filling up the [cruds object](https://github.com/owncloud/calendar-rework/pull/73/files#diff-ecd0e6e30bcf39e2539bf6ddc6eca715R18) (we have to use ACL here, right?) or take care of implementing something like a SharingService. Would appreciate it!

btw, is there a reason why updating the supported components is not supported? I get a 403 when sending a proppatch

DeepDiver1975:
> btw, is there a reason why updating the supported components is not supported? I get a 403 when sending a proppatch

can you send me the request so that I can debug this?

georgehrke:
The request is a `PROPPATCH` to a calendar. In my case `http://owncloud.dev/remote.php/dav/calendars/admin/test-1/`

The payload looks like:
```
<d:propertyupdate xmlns:c="urn:ietf:params:xml:ns:caldav" xmlns:d="DAV:" xmlns:a="http://apple.com/ns/ical/" xmlns:o="http://owncloud.org/ns"><d:set><d:prop><c:supported-calendar-component-set><c:comp name="VEVENT"/><c:comp name="VJOURNAL"/></c:supported-calendar-component-set></d:prop></d:set></d:propertyupdate>
```

DeepDiver1975:
thx - let me test this

DeepDiver1975:
debugging .....

DeepDiver1975:
@georgehrke supported-calendar-component-set is defined as protected property within the caldav plugin of sabre

@evert mind sharing the background ? THX

evert:
It's explicitly forbidden by the specification to change the calendar type after creation. What happens for instance if you remove VTODO from supported-calendar-component-set that already has `VTODOs` in it.

evert:
See here: http://tools.ietf.org/html/rfc4791#section-5.2.3

DeepDiver1975:
THX @evert

DeepDiver1975:
@georgehrke I'm seeing this on load of the application

````
Error: [$injector:unpr] Unknown provider: CalendarModelProvider <- CalendarModel <- DatepickerController
http://errors.angularjs.org/1.4.8/$injector/unpr?p0=CalendarModelProvider%20%3C-%20CalendarModel%20%3C-%20DatepickerController
    at angular.js?v=d473b72418bde628631f82dec41d21dc:68
    at angular.js?v=d473b72418bde628631f82dec41d21dc:4334
    at Object.getService [as get] (angular.js?v=d473b72418bde628631f82dec41d21dc:4482)
    at angular.js?v=d473b72418bde628631f82dec41d21dc:4339
    at getService (angular.js?v=d473b72418bde628631f82dec41d21dc:4482)
    at Object.invoke (angular.js?v=d473b72418bde628631f82dec41d21dc:4514)
    at extend.instance (angular.js?v=d473b72418bde628631f82dec41d21dc:9182)
    at nodeLinkFn (angular.js?v=d473b72418bde628631f82dec41d21dc:8299)
    at angular.js?v=d473b72418bde628631f82dec41d21dc:8572
    at processQueue (angular.js?v=d473b72418bde628631f82dec41d21dc:14792)(anonymous function) @ angular.js?v=d473b72418bde628631f82dec41d21dc:12520(anonymous function) @ angular.js?v=d473b72418bde628631f82dec41d21dc:9292processQueue @ angular.js?v=d473b72418bde628631f82dec41d21dc:14800(anonymous function) @ angular.js?v=d473b72418bde628631f82dec41d21dc:14808Scope.$eval @ angular.js?v=d473b72418bde628631f82dec41d21dc:16052Scope.$digest @ angular.js?v=d473b72418bde628631f82dec41d21dc:15870Scope.$apply @ angular.js?v=d473b72418bde628631f82dec41d21dc:16160bootstrapApply @ angular.js?v=d473b72418bde628631f82dec41d21dc:1679invoke @ angular.js?v=d473b72418bde628631f82dec41d21dc:4523doBootstrap @ angular.js?v=d473b72418bde628631f82dec41d21dc:1677bootstrap @ angular.js?v=d473b72418bde628631f82dec41d21dc:1697angularInit @ angular.js?v=d473b72418bde628631f82dec41d21dc:1591(anonymous function) @ angular.js?v=d473b72418bde628631f82dec41d21dc:29013c @ jquery.min.js?v=d473b72418bde628631f82dec41d21dc:4p.fireWith @ jquery.min.js?v=d473b72418bde628631f82dec41d21dc:4x.extend.ready @ jquery.min.js?v=d473b72418bde628631f82dec41d21dc:4q @ jquery.min.js?v=d473b72418bde628631f82dec41d21dc:4

**Comment by georgehrke** [comment]
> @georgehrke I'm seeing this on load of the application

Yes, don't ask ... 
Will be fixed once this is merged and we rebased it.
https://github.com/owncloud/calendar-rework/pull/80

**Comment by DeepDiver1975** [comment]
ok ... what about saving an event?

````
Object {kind: "timezone_missing", missing_timezone: undefined}(anonymous function) @ angular.js?v=d473b72418bde628631f82dec41d21dc:12520(anonymous function) @ angular.js?v=d473b72418bde628631f82dec41d21dc:9292Scope.$apply @ angular.js?v=d473b72418bde628631f82dec41d21dc:16157(anonymous function) @ angular.js?v=d473b72418bde628631f82dec41d21dc:23618x.event.dispatch @ jquery.min.js?v=d473b72418bde628631f82dec41d21dc:5v.handle @ jquery.min.js?v=d473b72418bde628631f82dec41d21dc:5

**Comment by DeepDiver1975** [comment]
> Will be fixed once this is merged and we rebased it.
> #80

so we better push #80 ????

**Comment by georgehrke** [comment]
> ok ... what about saving an event?

We need the TimezoneService for that. We have the raw ics for all olsen timezones on the server, the timezone needs to be queried and appended to the ics.
Would be good if this timezone service wouldn't need to call php, because that would add some server-side dependency for vobject. It should rather get the ics files directly, they are inside `/timezones/`. Maybe @tomneedham is interested in taking care of the timezone-service.

**Comment by DeepDiver1975** [comment]
rebasing after merge of #80

**Comment by DeepDiver1975** [comment]
rebased

**Comment by DeepDiver1975** [comment]
> Would be great if you could take care of filling up the cruds object (we have to use ACL here, right?)

Okay - we can ask for the acl in the propfind - no problem - but the dav client seems to have issues in parsing the response .....

... let's see if I can fix this.

**Comment by DeepDiver1975** [comment]
> ) or take care of implementing something like a SharingService.

I need to bring in sharing to calendar ...

**Comment by georgehrke** [comment]
I'm working on the EventsService and the etag returned by the server contains quotes. Is that intended?
Seems strange to me:

Request Header:
```
REPORT http://owncloud.dev/remote.php/dav/calendars/admin/test-1/
Depth: 1
Content-Type: application/xml; charset=utf-8
```

Request Payload:
```
<c:calendar-query xmlns:c="urn:ietf:params:xml:ns:caldav" xmlns:d="DAV:" xmlns:a="http://apple.com/ns/ical/" xmlns:o="http://owncloud.org/ns"><d:prop><d:getetag/><c:calendar-data/></d:prop><c:filter><c:comp-filter name="VCALENDAR"><c:comp-filter name="VEVENT"><c:time-range start="20151206T141436Z" end="20151220T141436Z"/></c:comp-filter></c:comp-filter></c:filter></c:calendar-query>
```

Response:
```
<?xml version="1.0"?>
<d:multistatus xmlns:d="DAV:" xmlns:s="http://sabredav.org/ns" xmlns:cal="urn:ietf:params:xml:ns:caldav" xmlns:cs="http://calendarserver.org/ns/" xmlns:card="urn:ietf:params:xml:ns:carddav" xmlns:oc="http://owncloud.org/ns">
 <d:response>
  <d:href>/remote.php/dav/calendars/admin/test-1/0F9D1390-C380-466F-8110-4B413B3C2319.ics</d:href>
  <d:propstat>
   <d:prop>
    <d:getetag>&quot;a75740f3096a2589500897e5261d6010&quot;</d:getetag>
    <cal:calendar-data>BEGIN:VCALENDAR
...
END:VCALENDAR
</cal:calendar-data>
   </d:prop>
   <d:status>HTTP/1.1 200 OK</d:status>
  </d:propstat>
 </d:response>
 <d:response>
  <d:href>/remote.php/dav/calendars/admin/test-1/CA3A2272-9F07-43B5-B2EF-AB5EA1D0562B.ics</d:href>
  <d:propstat>
   <d:prop>
    <d:getetag>&quot;abd5424e2b26dbd4604ffed2316bba24&quot;</d:getetag>
    <cal:calendar-data>BEGIN:VCALENDAR
...
END:VCALENDAR
</cal:calendar-data>
   </d:prop>
   <d:status>HTTP/1.1 200 OK</d:status>
  </d:propstat>
 </d:response>
</d:multistatus>

```

georgehrke:
It's the same [here](http://sabre.io/dav/building-a-caldav-client/), so it's probably all right, never mind then

DeepDiver1975:
> I'm working on the EventsService and the etag returned by the server contains quotes. Is that intended?

Our own implementation follows the reference pdo implementation by @evert - so I assume it is correct

DeepDiver1975:
> It should rather get the ics files directly, they are inside /timezones/

but I guess we better have them as jCal?

Or maybe even as plain js?

georgehrke:
> but I guess we better have them as jCal?

I don't think that would be of great help, because the client supports both. It wouldn't hurt either.

> Or maybe even as plain js?

Not sure about this, I think I would rather stick with ical or jcal. We don't have to parse all timezones, but just a few. So the performance benefits of using plain js would be rather small.

georgehrke:
This is ready to be merged imho.
Everything that works in master works in this branch as well.
Stuff like creating events is also not working in master and can be implemented after this was merged.

raghunayyar:
:+1:

Issue #76, opened 2015-12-10
Left navigation doesn’t collapse on mobile


jancborchardt:
For some reason the normal app navigation doesn’t collapse on mobile. The toggle shows but the navigation also still stays.

Anything you do differently from the default app HTML?

raghunayyar:
@jancborchardt self assigning. I will check this over the weekend.

georgehrke:
The question is how we want to deal with mobile anyway?

afaik full calendar is barely working on mobile, so just show a more or less working full calendar or rather some instructions on how to setup caldav on your mobile device (with an option to proceed to full calendar)?

jancborchardt:
I'd say it works fine, or at least good enough that we should show the actual interface instead of only instructions to set up Caldav.

Keep in mind mobile is not only very small smartphone screens but also tablets etc. If it doesn't work, switching to the Agenda view will do.

PR #77, opened 2015-12-11
Fixes CSS to support SnapJS from core.

raghunayyar:
A fix for #76
@georgehrke @jancborchardt please check.


georgehrke:
please rebase

PR #79, opened 2015-12-12
Makes App Nav consistent with core.

raghunayyar:
Removed a lot of our styles to match the app navigation of calendar with files. cc @jancborchardt @georgehrke


georgehrke:
:+1:

PR #84, opened 2015-12-17
Only build master branch on push on travis

DeepDiver1975:


DeepDiver1975:
So - there is only one travis job remaining in the check list above -> merge

PR #89, opened 2015-12-20
Fixes Calendar hover row colour issue.

raghunayyar:
Fixes https://github.com/owncloud/calendar-rework/issues/86
@georgehrke

### DIFF
--- a/css/calendar.css
+++ b/css/calendar.css
@@ -23,8 +23,12 @@
 
 
 #app-content.loading {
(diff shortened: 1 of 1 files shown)


georgehrke:
When selecting a few cells, the first row is colored grey.
This is different from http://fullcalendar.io/js/fullcalendar-2.5.0/demos/selectable.html,
probably some core styles as well

PR #91, opened 2015-12-22
Timezone service

georgehrke:
please review @raghunayyar


DeepDiver1975:
Wasn't there the idea to load the timezone files static?

raghunayyar:
It works for me :+1:.

georgehrke:
> Wasn't there the idea to load the timezone files static?

My goal was to remove the server side dependency of vobject. The current solution is just piping thru the file and not parsing and serializing it, like it did before.

PR #97, opened 2015-12-23
WIP: Datepicker fix

georgehrke:
@raghunayyar please properly fix the alignment of the buttons with the ones in the row below


raghunayyar:
Can't get more exact than this now. @georgehrke please check.

georgehrke:
looks much better now, thx :)

PR #100, opened 2015-12-24
make scrutinizer happy be renaming getName to smth else

georgehrke:


raghunayyar:
What was the issue otherwise?

PR #103, opened 2015-12-24
improve datepicker behaviour

georgehrke:
please review @raghunayyar


raghunayyar:
I dont like the format of the date, can we use a format analogous to `26th December, 2015`?

PR #110, opened 2016-01-08
WIP: Migrate import to dav

georgehrke:
Using the dav backend requires some changes to the import.
Previously there was an REST API endpoint to which one could send an ics with multiple objects.
This doesn't work with dav. We have to import objects one at a time.

The newly introduced `SplitterService` is taking care of splitting up the big ics blobs.

fyi @raghunayyar

### DIFF
--- a/appinfo/application.php
+++ b/appinfo/application.php
@@ -50,8 +50,10 @@ class Application extends App {
 		});
 		$container->registerService('ViewController', function(IAppContainer $c) {
 			$request = $c->query('Request');
+			$userSession = $c->getServer()->getUserSession();
+			$config = $c->getServer()->getConfig();
 
-			return new Controller\ViewController($c->getAppName(), $request);
+			return new Controller\ViewController($c->getAppName(), $request, $userSession, $config);
 		});
 	}
 
--- a/controller/viewcontroller.php
+++ b/controller/viewcontroller.php
@@ -29,8 +29,34 @@ use OCP\AppFramework\Http\DataDisplayResponse;
 use OCP\AppFramework\Http\JSONResponse;
 use OCP\AppFramework\Http\NotFoundResponse;
 use OCP\AppFramework\Http\TemplateResponse;
+use OCP\IConfig;
+use OCP\IRequest;
+use OCP\IUserSession;
 
 class ViewController extends Controller {
+
+	/**
+	 * @var IConfig
+	 */
+	private $config;
+
+	/**
+	 * @var IUserSession
+	 */
+	private $userSession;
+
+	/**
+	 * @param string $appName
+	 * @param IRequest $request an instance of the request
+	 * @param IConfig $config
+	 * @param IUserSession $userSession
+	 */
+	public function __construct($appName, IRequest $request,
+								IUserSession $userSession, IConfig $config) {
+		parent::__construct($appName, $request);
+		$this->config = $config;
+		$this->userSession = $userSession;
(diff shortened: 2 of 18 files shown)


georgehrke:
There is one remaining issue I would like to resolve:

ICS files can contain events, todos and journals. Calendars may not support all three types.
So the question is what to do:
- If e.g. a user tries to import an ics with events and todos:
 - 1. only show calendars that support at least events and todos (if none, there is still the new calendar option)
 - 2. show all calendars but add a warning that not all objects from the ics will be imported

raghunayyar:
@georgehrke can we show what precisely cant be imported?

georgehrke:
theoretically yes, but that means implementing some parsing logic for todos and journals, which will be quite some work.

another issue is how to notify the user about events (or todos or journals) rejected by the server.

raghunayyar:
cc @jancborchardt

karlitschek:
awesome work guys. :-)
do you have concrete plans already how the migration story for people moving from owncloud 8.2 to 9.0 looks like?
idealy migration is fully automatic. so the new calendar or maybe owncloud core detects that old calendar db tables exist and automatically moves data over without manual work
what do you think?
@DeepDiver1975

georgehrke:
> do you have concrete plans already how the migration story for people moving from owncloud 8.2 to 9.0 looks like?

I agree that it should ideally be done fully automatic. AFAIK the calendar table is pretty much the same, but the objects table is a bit different. It now contains firstoccurrence and lastoccurrence, so we have to parse every single object :/

So maybe some extra upgrade routine after the actual update to ownCloud 9.

karlitschek:
Works for me as long as this additional upgrade routine is executed automatically without a manual step by the user.
What do you think?

georgehrke:
Hm, maybe an additional step for the admin?

karlitschek:
well. the goal ist that users upgrade from 8.2 to 9.0 and everything works without any manual work and configuration change and manual script execution or documentation reading.
This is the goal.

DeepDiver1975:
Calendar and contacts migration is a core task. I'm working on a solution. Wip PR is open in core.

georgehrke:
> Wip PR is open in core.

Can you point us there? :)

DeepDiver1975:
> Can you point us there? :)

https://github.com/owncloud/core/pull/21333

karlitschek:
@DeepDiver1975 Awesome :-)

georgehrke:
@jancborchardt any opinion? :)

jancborchardt:
@georgehrke see the discussion in the core issue. I agree with @karlitschek that it's important to be as automatic as possible. However we need to gauge to which point this is possible.

georgehrke:
@jancborchardt I mean the actual issue mentioned in the top of this issue ;)

https://github.com/owncloud/calendar-rework/pull/110#issuecomment-170279268

jancborchardt:
I'd go for option 2 as it provides proper feedback.

georgehrke:
Okay, will go for that :)

georgehrke:
@raghunayyar please check

DeepDiver1975:
Tested import of a single event :+1:

DeepDiver1975:
bulk import works as well :+1: 

merge from my pov

DeepDiver1975:
hmm .. bloody php builtin webserver doesnot support MKCALENDAR :imp:

PR #118, opened 2016-01-25
get rid of database.xml

georgehrke:
This app doesn't require any database. No need to create tables used by legacy versions on new installs ;)


DeepDiver1975:
:+1:

PR #122, opened 2016-01-26
CalendarSharing

georgehrke:


js/public/app.js
```diff
@@ -1473,8 +1476,11 @@ app.factory('Calendar', ['$rootScope', '$filter', 'VEventService', 'TimezoneServ
 			this._properties.order = order;
 			this._setUpdated('order');
 		},
-		get cruds() {
-			return this._properties.cruds;
+		get writable() {
+			return this._properties.writable;
+		},
+		get shareable() {
+			return this._properties.shareable;
```
tomneedham: We're getting rid of this right?

js/public/app.js
```diff
@@ -1351,12 +1349,11 @@ app.factory('Calendar', ['$rootScope', '$filter', 'VEventService', 'TimezoneServ
...
 					vtodo: false
 				},
-				cruds: {
-					create: props.canWrite,
-					read: true,
-					update: props.canWrite,
-					delete: props.canWrite,
-					share: props.canWrite
+				writable: props.canWrite,
+				shareable: props.canWrite,
```
tomneedham: And this

js/app/service/fchelper.js
```diff
@@ -79,7 +79,7 @@
 	 */
 	function addCalendarDataToFCData(fcData, calendar) {
 		fcData.calendar = calendar;
-		fcData.editable = calendar.cruds.update;
+		fcData.editable = calendar.editable;
```
tomneedham: Shoudl this be calendar.writeable?

js/public/app.js
georgehrke: No, a CalDAV server can disallow sharing for certain calendars. 
https://trac.calendarserver.org/browser/CalendarServer/trunk/doc/Extensions/caldav-sharing.txt#L704

We are just getting rid of the can re share option, because it's giving a false sense of security.

js/app/service/fchelper.js
georgehrke: Yes, indeed.
Will fix later today

js/public/app.js
tomneedham: Great, got it.

js/app/service/calendarservice.js
```diff
@@ -24,6 +24,9 @@
 app.service('CalendarService', ['DavClient', 'Calendar', function(DavClient, Calendar){
 	'use strict';
 
+	this.SHARING_USER = 0;
+	this.SHARING_GROUP = 1;
```
tomneedham: Could we use: `OC.Share.SHARE_TYPE_USER` and `OC.Share.SHARE_TYPE_GROUP` maybe?

georgehrke: sure, will change it :)

dominicrico:
@tomneedham how can I help you?

tomneedham:
@georgehrke Is the CalendarService stuff done? From the UI perspective its just wiring it all up. Plus we need to make a plan with @DeepDiver1975 about user/group discovery. @georgehrke Did you see https://github.com/owncloud/core/pull/21964

DeepDiver1975:
> lus we need to make a plan with @DeepDiver1975 about user/group discovery.

let's discuss this on irc

DeepDiver1975:
@georgehrke @tomneedham calendar sharing was just merged into master - FYI

tomneedham:
Awesome. I'll be on this this afternoon. @DeepDiver1975 What was the result of our 'conversation' on IRC regarding sharee discovery? Hack files_sharing for now? Are there cases when that endpoint wont be available?

DeepDiver1975:
> Hack files_sharing for now?

yes - for now

tomneedham:
Issue: 
- [ ] Sharing icon doesn't appear on safari right now

tomneedham:
So right now we have a nice discovery box, searching users and groups using the files_sharing endpoint. On select it calls calendar.share with the relevant details and create a share which is by default not writeable (like in files). There is a simple checkbox to change the write permissions.

screen shot 2016-02-04 at 18 03 35
screen shot 2016-02-04 at 18 03 28

@georgehrke What needs to be done (if anything) in the calendar service here to wire this up to @DeepDiver1975's sharing stuff now in core?

tomneedham:
Rebased

georgehrke:
> @georgehrke What needs to be done (if anything) in the calendar service here to wire this up to @DeepDiver1975's sharing stuff now in core?

Some XML voodoo, will take care

tomneedham:
> Some XML voodoo, will take care

Cool. The UI is wired up to the calendar model which is about where my angular knowledge ends at the moment. Will it handle the data binding with the editable checkboxes being changed for example?

georgehrke:
> Will it handle the data binding with the editable checkboxes being changed for example?

I think so (will check), but we also need to send a new XML request, when the writable property changes. will take care of this as well.

tomneedham:
Awesome

georgehrke:
@DeepDiver1975 Can you tell me what I'm doing wrong? I just can't seem to find my mistake :confused: 
https://gist.github.com/georgehrke/8cea0922c75530c07dc0

DeepDiver1975:
@georgehrke the xml payload misses the <xml tag ???

georgehrke:
Tried that as well, doesn't work

DeepDiver1975:
the csrf token is missing - which is requires since you are not using basic auth

georgehrke:
adding the csrf token worked, thx a lot :)

georgehrke:
Why don't the other caldav calls don't require the csrf token?

DeepDiver1975:
> Why don't the other caldav calls don't require the csrf token?

Only POST requires csrf - non of the other verbs used for webdav are vulnerable to csrf

georgehrke:
I forgot one thing

DeepDiver1975:
Really nice! I love it! :+1:

tomneedham:
IMO, todo:
 - [ ] Fix jerk action when searching for typeahead (style the loading element)
 - [ ] Close the sharee editor when click away
 - [ ] Move cursor to input on share icon click for usabilitiy
 - [ ] BUG: Sharee sees themself in list of sharees. Do we disable sharing for calendar recipients?

PR #126, opened 2016-01-31
Updated Grunt and fixed lint errors

dominicrico:
Because Grunt was always failing and I'm not the --force guy, because this is not helpful in my opinion, I removed the linting errors by adding returns in the if statements and also added a reporter to the jshint task which displays the filename of the failed file.


georgehrke:
Thanks for your contribution :)

Will review it tomorrow, looking forward to seeing more contributions from you ;)
I will also look for junior jobs and mention you there

georgehrke:
> added dependencie batch for the npm dependencies into read

Can you add one for bower as well? :)

dominicrico:
The thing is that I have found out, that the badge will always display "unknown" because the package.json is not in the root folder...

dominicrico:
There are three different services for that type of badges. 
- [Gemnasium](http://gemnasium.com)
- [VersionEye](https://www.versioneye.com)
- [David](https://david-dm.org/)

The thing about David is that they only support NPM deps, no bower deps and the package.json must be in the root folder of the repo. The others support both, but one of the owners must register it self there to grand one of the services the right to read the package.json, which maybe needs to be in the root folder (but I dont really know).

georgehrke:
Looks good, but please remove the david-dm.org link for now :)
Will merge afterwords

georgehrke:
@dominicrico If you look for further stuf to work on, you might collaborate with @tomneedham on https://github.com/owncloud/calendar-rework/pull/122 :)

dominicrico:
Yes, of course! Thanks @georgehrke.

PR #183, opened 2016-02-12
rename alarm to reminders

georgehrke:
fixes https://github.com/owncloud/calendar-rework/issues/167


raghunayyar:
Is this the only place where the replacement needs to be done? I thought there are more cases where we have used alarms. Will confirm once and get this in after that.

georgehrke:
Yes, we don't need to rename the variables ;)

PR #185, opened 2016-02-12
Overall timezone display fixes

### LINKED ISSUE #154: When disabling allday and timezone is null, set it to defaulttimezone

### LINKED ISSUE #155: Show timezone in popover
If the timezone differs from defaulttimezone, show the actual timezone of the event.
This doesn't need to be a select like in the sidebar, just a span that displays the selected one.

### LINKED ISSUE #156: show timezone when opening sidebar if selected one is not defaulttimezone

### LINKED ISSUE #160: group timezones by continent and replace _ with space in displayname

### LINKED ISSUE #171: add timezone filter to display timezone in a more user-friendly way
e.g. New York (America) instead of America/New_York

georgehrke:
fixes #171
fixes #160 
fixes #156 
fixes #155 
fixes #154


georgehrke:
rebased

DeepDiver1975:
bildschirmfoto von 2016-02-16 11-25-51

bildschirmfoto von 2016-02-16 11-27-21

bildschirmfoto von 2016-02-16 11-27-43

bildschirmfoto von 2016-02-16 11-28-10

bildschirmfoto von 2016-02-16 11-28-33

georgehrke:
@DeepDiver1975 What are you trying to tell me? :)

DeepDiver1975:
bildschirmfoto von 2016-02-16 11-29-09

bildschirmfoto von 2016-02-16 11-29-24

bildschirmfoto von 2016-02-16 11-30-01
bildschirmfoto von 2016-02-16 11-30-27
bildschirmfoto von 2016-02-16 11-30-46

georgehrke:
- [ ] remove second underscore

DeepDiver1975:
And there is a dot missing after all these St ...

georgehrke:
> not really a name

It actually is ;)
https://en.wikipedia.org/wiki/Longyearbyen

georgehrke:
- [ ] remove dashes from name

georgehrke:
- [ ] replace `St` with `St.`

georgehrke:
> too much sashes
bildschirmfoto von 2016-02-16 11-29-09

It's actually called Port-au-Prince
https://en.wikipedia.org/wiki/Port-au-Prince

DeepDiver1975:
> It actually is ;)
> https://en.wikipedia.org/wiki/Longyearbyen

haha - looked like 'Long year by en'

DeepDiver1975:
> It's actually called Port-au-Prince

but then we have a few whitespaces too much - right?

georgehrke:
> but then we have a few whitespaces too much - right?

- [ ] I will check

DeepDiver1975:
let me play with this during lunch time :wink:

georgehrke:
sure :)

DeepDiver1975:
@raghunayyar @georgehrke Why is Porto-au-Prince in this list with slashes and not with dashes?
https://github.com/owncloud/calendar-rework/blob/master/js/app/service/timezonelistprovider.js#L197

There is basically no way to differentiate 

```js
			'America\/North_Dakota\/New_Salem',
			'America\/Port\/Au\/Prince',

georgehrke:
> Why is Porto-au-Prince in this list with slashes and not with dashes?

That's a bug, I extracted the timezone names from the file names in timezones/.
We definitely need a better solution for this in the future. Like greping for the line with `TZID:...`

georgehrke:
Actually we should validate that list, to make sure all entries map to a proper timezone.
Should be fairly easy with a test.
Let me open a ticket for that, nothing we have to fix before alpha2

DeepDiver1975:
I assume there is some mapping in place to get the proper timezone ics file for a given timezone in that list. Where is that happening?

georgehrke:
https://github.com/owncloud/calendar-rework/blob/master/controller/viewcontroller.php#L94

DeepDiver1975:
Okay - porto-au-prince is broken - it cannot load the ics file

georgehrke:
because in the name it's Au, it should be au.

DeepDiver1975:
> because in the name it's Au, it should be au.

with the slashes in the name it was trying to load 
http://localhost:8888/index.php/apps/calendar/v1/timezones/AMERICA/PORT/AU/PRINCE.ics

DeepDiver1975:
@georgehrke @raghunayyar are you okay with my changes?

georgehrke:
will review and merge afterwords

PR #198, opened 2016-02-14
improve popover position in agenda views

georgehrke:
fixes https://github.com/owncloud/calendar-rework/issues/191


DeepDiver1975:
:+1:

PR #199, opened 2016-02-14
add temporary fcEvent while creating event

georgehrke:
fixes https://github.com/owncloud/calendar-rework/issues/172


DeepDiver1975:
@georgehrke how can I test this? THX

georgehrke:
Go to month view, instead of the blue selection (like in master) there should be an actual event when creating new events

DeepDiver1975:
gotcha! :+1:

PR #209, opened 2016-02-17
fix calendar not disappearing in calendar editor after deletion

georgehrke:
fixes https://github.com/owncloud/calendar-rework/issues/157


DeepDiver1975:
:+1:

PR #210, opened 2016-02-17
ng-cloak entire CalendarListController in order to prevent flashing

georgehrke:
fixes https://github.com/owncloud/calendar-rework/issues/152


DeepDiver1975:
:+1:

PR #219, opened 2016-02-18
use fullcalendar locales

georgehrke:
fixes https://github.com/owncloud/calendar-rework/issues/217

please review @enoch85 
(note that you can't apply this patch to alpha2, because the lang-all.js is missing)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -45,7 +45,6 @@ appstore: clean
 	--exclude=$(project_dir)/js/vendor/**/karma.conf.js \
 	--exclude=$(project_dir)/js/vendor/davclient.js/index.html \
 	--exclude=$(project_dir)/js/vendor/fullcalendar/dist/gcal.js \
-	--exclude=$(project_dir)/js/vendor/fullcalendar/dist/lang-all.js \
 	--exclude=$(project_dir)/js/vendor/fullcalendar/dist/lang \
 	--exclude=$(project_dir)/js/vendor/ical.js/build/benchmark \
 	--exclude=$(project_dir)/js/vendor/ical.js/lib \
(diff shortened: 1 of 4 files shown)


enoch85:
A did a git clone and edited the code directly, set correct permissions and tested.

Now I can't add specific time it's always a whole day event, and when I tested to add events on two different days and then pressed cancel - both events disappeared. Maybe I'm doing it wrong?

georgehrke:
What code did you edit?
If you perform a git clone, there is no need to change any code.

enoch85:
@georgehrke The code isn't merged yet, and I cloned masted.
I changed the files as you did accordingly.

georgehrke:
just try to checkout the branch `use_fullcalendar_locales`

enoch85:
Aah, nice. Will try.

enoch85:
No still same issue here: :-1:

enoch85:
Can someone else please confirm? @jancborchardt maybe?

DeepDiver1975:
Works for me as far as I can tell - but I would expect 4:00 instead of 4 as hour label

bildschirmfoto von 2016-02-19 09-01-39

georgehrke:
I see your point, but having 04 instead of 04:00 is still better than having allday instead of ganztägig and 2/15 instead of Di. 15.2

I don't see something small we can do about the issue with 04. We are using the full calendar locales and  patching those is not a solution. I have an idea what to do about this, but that's nothing we can do in time for ownCloud 9.

My point is: This is not perfect, but still good progress, so please get this in ;)

@enoch85 What language did you select in the personal settings. Maybe it happens only with your locale.

DeepDiver1975:
> My point is: This is not perfect, but still good progress, so please get this in ;)

sure :+1:

enoch85:
@georgehrke Swedish. I say let's get it in, and we can have a look at it later. (y)

PR #229, opened 2016-02-20
add organizer when event has attendees

georgehrke:
fixes https://github.com/owncloud/calendar-rework/issues/195


DeepDiver1975:
tested :+1:

PR #240, opened 2016-02-26
don't offer edit option for shared calendars

georgehrke:
fixes https://github.com/owncloud/calendar-rework/issues/220


georgehrke:
@raghunayyar You can just merge it, if it's okay. I don't think the scrutinizer test will ever finish

Issue #112, opened 2016-01-09
versioning scheme and maintenance periods


georgehrke:
I would like to get a few opinions on the following topics:
- Which version scheme should we use? 
  - `major.maintenance` or `major.minor.maintenance`
- Keep our major release schedule in sync with ownCloud or do something completely independent?
  - By relying on the standardized CalDAV interface and mostly javascript I don't expect too many breaking changes with major ownCloud updates. Though major updates of ownCloud almost always introduced design changes that required action on our side.
  - Can we handle major ownCloud updates with maintenance updates or do we need at least a major (or minor) update.
- How long do we want to support major (or minor) versions of the calendar? Do we want to support multiple major ownCloud versions or only the latest? If yes, is there a need to release different versions for different ownCloud versions? (related to question two)

MorrisJobke:
I would propose a separate version schema. With the music app I always tried to maintain exactly one version of the app that works for all supported OC versions. I added some fixes, that went later into core directly to support different versions, but it's far easier to maintain one code base rather than maintain multiple.

All the news app has gone a similar way. Except that supported versions are reduced a lot and @BernhardPosselt pointed out that an older version could be used.

And it isn't that hard to maintain a stable owncloud app. That's why we build the OCP API ;) And you use the CalDAV API which is even better specified and more stable :)

BernhardPosselt:
IMHO: Newest release supports newest ownCloud version and further versions if its trivial. 

If it's not trivial let someone pay you for backporting security fixes, that's it ;)

DeepDiver1975:
With the mail app we aim to have one app which is compatible with all supported oc server versions.
With the new calendar app this will of course be starting with oc9.

georgehrke:
Thanks for all the valuable feedback :)

So we will go with: 
- `major.maintenance`
- only one supported version of the app that's supposed to cover all supported ownCloud versions.
- do something independent from the ownCloud release cycle but obviously release updates prior to major ownCloud updates if needed.

I'll also document this in the wiki

PR #256, opened 2016-03-02
Always test stable branches

nickvergessen:


georgehrke:
GRMPF, there are not supposed to be stable branches :confused:

georgehrke:
@karlitschek @cmonteroluque Please remove the calendar from the @owncloud-packager script :)

nickvergessen:
why not? those are really helpful, or do you really want one branch that works with all versions?

georgehrke:
yes, that's also the way it works for news and mail

georgehrke:
We discussed this here: https://github.com/owncloud/calendar/issues/112

karlitschek:
i removed it from the branching script.

georgehrke:
thx :)

ghost:
stable7 build script fixed to work with archive repositories for calendar and contacts

PR #257, opened 2016-03-02
Update the env.global.CORE_BRANCH setting in .travis.yml

nickvergessen:


georgehrke:
calendar is not part of the release, there shouldn't be stable branches in the first place

PR #278, opened 2016-03-07
add rudimentary support for editing recurring events

georgehrke:


raghunayyar:
The code looks good :+1:

PR #279, opened 2016-03-07
hide class input for 1.0

georgehrke:
Core ignores the class property at the moment. We should hide it, because it's confusing


raghunayyar:
:+1:

PR #280, opened 2016-03-07
show direct caldav link as title

georgehrke:
thunderbird users need the entire caldav link for each calendar


raghunayyar:
:+1:

PR #296, opened 2016-03-08
Design fixes

jancborchardt:
Lots of design fixes all over the place! Check the commits and test drive it. Please review @owncloud/designers @raghunayyar @georgehrke @enoch85 @MeCias

Sorry for not doing this earlier – was swamped with Contacts. :)

### DIFF
--- a/css/calendarlist.css
+++ b/css/calendarlist.css
@@ -250,3 +250,10 @@ ul.dropdown-menu  li {
 ul.dropdown-menu li > a:hover {
 	background-color: grey !important;
 }
+
+
+/* Fullcalendar modifications */
+
+.fc th {
+	opacity: .3;
+}
--- a/css/calendarlist.css
+++ b/css/calendarlist.css
@@ -24,7 +24,6 @@
 
 #app-navigation .calendar-list {
 	max-height: 400px;
-	overflow-y: scroll;
 }
 #app-navigation .app-navigation-input {
 	width: 95%;
--- a/css/calendarlist.css
+++ b/css/calendarlist.css
@@ -257,3 +257,13 @@ ul.dropdown-menu li > a:hover {
 .fc th {
 	opacity: .3;
 }
+.fc td.fc-widget-header {
+	border-top: none;
+}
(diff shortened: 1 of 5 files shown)


raimund-schluessler:
@jancborchardt Your fixes improved the design, but I still see unnecessary scrollbars. See my screenshots before:
before

and after:
after

The scrollbar on the far right scrolls the whole calendar view so that the headers and controls are not visible anymore. Only the settings button stays fixed. The browser is Firefox 44.0.2 on Windows 10.

jancborchardt:
:( I can’t reproduce that with any of my browsers and don’t have Windows. @georgehrke @raghunayyar any idea?

MeCias:
I can confirm the 2 scroll bars on the month view with Firefox 45 - 64 bit on Windows 10

Otherwise nothing to complain. looks very good. Thank you.

enoch85:
:+1: Not tested on Windows

georgehrke:
Can't reproduce the the scrollbar issue on OS X either. Will boot up my windows machine, but not before thursday afternoon

jancborchardt:
@georgehrke well, it’s not worse than the current style in that regard, so how about merging? It’s big enough as it is.

georgehrke:
There won't be a release before Thursday, so I don't see a need to hustle and not fix it carefully for all supported browsers :)

georgehrke:
@raimund-schluessler @MeCias Please check again :)

@jancborchardt When the list in the left sidebar is too long and you start scrolling, the entire view including the calendar in #app-content scrolls. Can you please take a look?

raimund-schluessler:
@georgehrke The scrollbars at the calendarlist and the calendar view are gone. I also see the issue that the complete app scrolls, when the sidebar content is to high.

jancborchardt:
@georgehrke I would really prefer this pull request to be merged before any other unrelated additional changes are done on top of it. The scrollbars issue is completely unrelated, let’s do it in a separate PR, otherwise it will only block this one.

georgehrke:
sure :)

PR #319, opened 2016-03-11
get caldav properties out of calendar model

### LINKED ISSUE #314: Edit button is not shown for own calendars
### Steps to reproduce
1. Install OC 9.0.0 and latest calendar master
2. Create a new calendar or have a look at an existing one
3. Hover calendar, edit button should be shown
### Expected behaviour

Edit button does not show up
### Actual behaviour

Edit button does not show up (only share, download and delete are visible, see screenshot)
### Server configuration

Current Debian running on Virtual box

**Web server:**
Apache 2.4.10

**Database:**
Maria DB 10.0.22-MariaDB-0+deb8u1 - (Debian)

**PHP version:**
5.5 I guess

**ownCloud version:** (see ownCloud admin page)
OC 9.0.0

**Calendar version:** (see ownCloud apps page)
Current master e3b5be1

**Updated from an older ownCloud or fresh install:**
Fresh

Nothing significant in the logs.

The reason for this is, that `calendar.owner` is not set since at https://github.com/owncloud/calendar/blob/master/js/app/models/calendarmodel.js#L114 `owner` equals `/owncloud/remote.php/dav/principals/users/raimund` which does not fulfill `owner.startsWith('/remote.php/dav/principals/users/')` at https://github.com/owncloud/calendar/blob/master/js/app/models/calendarmodel.js#L115

I hope this helps.

### LINKED ISSUE #316: remove caldav property names from calendarmodel

georgehrke:
fixes https://github.com/owncloud/calendar/issues/316
fixes https://github.com/owncloud/calendar/issues/314

### DIFF
--- a/js/app/controllers/calcontroller.js
+++ b/js/app/controllers/calcontroller.js
@@ -66,7 +66,7 @@ app.controller('CalController', ['$scope', '$rootScope', '$window', 'CalendarSer
 			});
 
 			angular.forEach(newCalendars, function(calendar) {
-				calendar.registerEnabledCallback(function(enabled) {
+				calendar.registerCallback('enabled', function(enabled) {
 					if (enabled) {
 						showCalendar(calendar.url);
 					} else {
--- a/js/app/models/calendarmodel.js
+++ b/js/app/models/calendarmodel.js
@@ -1,61 +1,59 @@
 app.factory('Calendar', ['$rootScope', '$filter', 'VEventService', 'TimezoneService', 'RandomStringService', function($rootScope, $filter, VEventService, TimezoneService, RandomStringService) {
 	'use strict';
 
+	function generateTextColor(r,g,b) {
+		var brightness = (((r * 299) + (g * 587) + (b * 114)) / 1000);
+		return (brightness > 130) ? '#000000' : '#FAFAFA';
+	}
+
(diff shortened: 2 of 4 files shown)


georgehrke:
@raimund-schluessler should also fix #314

georgehrke:
wait, it doesn't at the moment :see_no_evil:

georgehrke:
it should now

raimund-schluessler:
Fixes #314 :+1:

raimund-schluessler:
Btw. what is it good for to "get caldav properties out of calendar model" ?

georgehrke:
There should be a distinct separation between the calendar Model and the CalendarService. The calendarModel should just be a calendarModel, it doesn't need to know that the property was once called `{http://apple.com/ns/ical/}calendar-color`.
Having all those namespace strings in the CalendarService and CalendarModel is duplicate code.
I would consider the solution in 1.0 bad design.

georgehrke:
This is not only for the sake of good quality code, but will also come very handy when we implement webcal. By then we will have two sources for calendars, the CalendarService and the yet to be implemented SubscriptionService.

raghunayyar:
Please rebase and merge. This is long due.

PR #367, opened 2016-03-24
move alldayevent checkbox above day/time picker

9R:
Move alldayevent checkbox above start end day and time picker

This way, the ui elements are in a better arrangement when creating / editing events using the TAB key to advance from one field to the next.

Currently the user has to TAB to the alldayevent checkbox, skipping the disabled time picker, then SHIFT-TAB back to add a start and end time.

### DIFF
--- a/templates/editor.popover.php
+++ b/templates/editor.popover.php
@@ -16,6 +16,14 @@
 		</fieldset>
 
 		<fieldset class="event-time events--fieldset" ng-disabled="readOnly">
+			<div class="events--checkbox pull-left">
+				<input type="checkbox" name="alldayeventcheckbox"
+					   ng-model="properties.allDay"
+					   id="alldayeventcheckbox" class="event-checkbox"
+					   ng-change="toggledAllDay()" />
+				<label for="alldayeventcheckbox"><?php p($l->t('All day Event'))?></label>
+			</div>
+			<div class="clear-both"></div>	
 			<div class="event-time-interior pull-left">
 				<span><?php p($l->t('starts')); ?></span>
 				<ocdatetimepicker ng-model="properties.dtstart.value" disabletime="properties.allDay"></ocdatetimepicker>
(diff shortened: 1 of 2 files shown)


georgehrke:
cc @jancborchardt

georgehrke:
there is also a tab index attribute. might be more appropriate here

9R:
@georgehrke: Thanks for the suggestion. I looked into using "tabindex" just now, but ran into some problems:

tabindex is already in use in another part of the site, namely in the banner at the top. This interferes with the desired order in the event editor if indices in both areas intersect. I was able to mitigate this by using reasonably high indices in the eventeditors but this still become a problem in the future if higher indices are introduces in banner elements.

Also if I understand it correctly the date and time input fields for start and end of an event are generated by the same js and I am not sure how to give those fields different indices.

edit: Nevermind the js problem. I just figured something out. I will submit a alternative.

Any hints how to better solve this?

jancborchardt:
Can you post a before/after screenshot so it’s easier to review UI changes? :)

9R:
@jancborchardt: The only UI change is the moved "All day Event"-check box. Here are some screenshots.

popover editor before change:
popover_before

popover editor after change:
popover_after

sidebar editor before change:
sidebar_before

sidebar editor after change:
sidebar_after

georgehrke:
Looks pretty misplaced imho, there are better ways to fix the tab index, like in the other pr you opened.

:-1: on this one from my side.
We should pursue https://github.com/owncloud/calendar/pull/368 imo

jancborchardt:
Yeah, :-1: from me as well. This checkbox directly correlates to enabling/disabling the time inputs, which it currently is directly next to. The change would move it out of place.

9R:
Ok. I'm not attached to this particular solution. It was just my first idea how to archive a better workflow when using ALT-TAB while creating new events.
Let's create an improved version of #368 as suggested by @georgehrke

PR #384, opened 2016-03-30
use ng-submit

### LINKED ISSUE #382: Pressing enter in new calendar input doesn’t create calendar
You need to manually click the button, which is quite annoying. Pressing enter after having put in a calendar name should also work.

(Observed when I watched @nickvergessen use the the Calendar app ;)

georgehrke:
fixes https://github.com/owncloud/calendar/issues/382

please review @raghunayyar @jancborchardt

### DIFF
--- a/js/app/controllers/calendarlistcontroller.js
+++ b/js/app/controllers/calendarlistcontroller.js
@@ -43,6 +43,7 @@ app.controller('CalendarListController', ['$scope', '$rootScope', '$window', 'Ca
 
 			$scope.newCalendarInputVal = '';
 			$scope.newCalendarColorVal = '';
+			angular.element('#new-calendar-button').click();
 		};
 
 		$scope.download = function (calendar) {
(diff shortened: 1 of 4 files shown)


BernhardPosselt:
Btw this should work out of the box when using forms, also check th ngSubmit directive: https://docs.angularjs.org/api/ng/directive/ngSubmit

BernhardPosselt:
Check this: https://github.com/owncloud/news/blob/master/templates/part.navigation.addfeed.php

TL;DR: bind to ng-submit and not ng-click for the submit button

georgehrke:
@BernhardPosselt Does ng-submit also catch the enter?

BernhardPosselt:
Yes, everything that you'd expect from a form. And it prevents page reload ;)

georgehrke:
okay, will try out, thx :)

raghunayyar:
:+1:

PR #387, opened 2016-03-30
cut of alpha value from hex rgba strings

### LINKED ISSUE #358: Cut off alpha value from #RRGGBBAA
I have set custom colors for different calendars using iCal. While they were shown with the old calendar (so they are saved in the database), the new Web UI displays all 'custom' colors as blue and also does not show the color indicator in the list view on the left (the square is always white, independent of the calendar visibility state).

Stock colors:
stock colors

Custom colors:
custom colors

georgehrke:
fixes https://github.com/owncloud/calendar/issues/358


raghunayyar:
:+1:

georgehrke:
rebased

georgehrke:
tests pass locally -> merge

PR #388, opened 2016-03-30
use proper user method to get email address

### LINKED ISSUE #292: use email address from user object
In the `div ui-calendar="uiConfig.calendar" id="fullcalendar"` there is an attribute: `data-emailaddress="jan@example.com"`

If anything, the correct email address should be taken from settings. :)

georgehrke:
fixes https://github.com/owncloud/calendar/issues/292


raghunayyar:
:+1:

PR #393, opened 2016-03-30
correct spacing in calendar list

jancborchardt:
Before (very cramped, not standard in ownCloud):
capture du 2016-03-30 14-33-52
After (same spacing as in Files, Mail and other apps):
capture du 2016-03-30 14-32-36

Please review @owncloud/designers @georgehrke

### DIFF
--- a/css/calendarlist.css
+++ b/css/calendarlist.css
@@ -56,7 +56,7 @@
 	display: block;
 	width: 100%;
 	line-height: 44px;
-	padding: 0 25px;
+	padding: 0 44px;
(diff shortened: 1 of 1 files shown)


jancborchardt:
(BTW, never ever mix top/left in CSS with margin-top and margin-left … they basically have the same effect. ;)

georgehrke:
:+1:

jancborchardt:
Rebased as well

georgehrke:
tests pass locally :+1:

PR #400, opened 2016-03-30
introduce more menu in calendar list

georgehrke:
ToDo:
- [x] show actions all the time

### DIFF
--- a/css/calendarlist.css
+++ b/css/calendarlist.css
@@ -122,7 +122,10 @@
 	padding: 0 5px 0 0;
 }
 #app-navigation .app-navigation-list-item .utils .action span {
-	opacity: 0.6;
+	opacity: .3;
+}
+#app-navigation .app-navigation-list-item .utils .action:hover span {
+	opacity: 1;
 }
 #app-navigation .app-navigation-list-item .active {
 	background: transparent !important;
@@ -137,7 +140,7 @@
 	/* Core Override */
 }
 #app-navigation .app-navigation-list-item > a {
-	padding: 0 44px !important;
+	padding: 0 66px 0 44px !important;
 }
 #app-navigation .app-navigation-list-item .editfieldset {
 	padding: 0 4px;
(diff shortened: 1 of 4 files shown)


georgehrke:
@jancborchardt I can't seem to figure out, why the actions only appear on hover. I can't find any CSS rules that would imply this. I'm probably missing the obvious :confused:

georgehrke:
@jancborchardt Should the icons (share and ...) be visible 24/7?

jancborchardt:
@georgehrke I’d say yes maybe let’s try to show them all the time. That’s why I also said that it might be good to put share in the dropdown as well. Then again, sharing is also needed as an indicator if it has been shared (as done in the Files app) so we kinda need it outside.

After this short introspective: Yeah let’s show them all the time, and use the share icon as indicator as well, just like in the Files app with the actions on each file row. :)

jancborchardt:
@georgehrke at the same time, please work on standardizing this menu with @PVince81 and @BernhardPosselt, documenting the markup at https://doc.owncloud.org/server/9.0/developer_manual/app/css.html#standard-layout

raghunayyar:
@georgehrke https://github.com/owncloud/calendar/blob/master/css/calendarlist.css#L120 if you give width: 120px or something over here, you can see the actions all the time. Let me know if it works. :)

georgehrke:
I figured out that it had something to do with the width, but no width being assigned at all was too obvious :see_no_evil:

georgehrke:
but assigning a width doesn't work for me. The actions are still 0x0 for some reason

georgehrke:
got it

raghunayyar:
Works for me. :+1:

georgehrke:
pass locally -> merge

jancborchardt:
@georgehrke can you document the markup then now at the docs page I referenced above? And work with @BernhardPosselt on standardizing it, in case it’s different from what’s done in the News app or Files app?

jancborchardt:
@georgehrke did the documentation of the markup happen, as mentioned above? cc @BernhardPosselt @owncloud/javascript

PR #402, opened 2016-03-30
fix app order

jancborchardt:
The idea is to have the sorting start with: Files, Activity, Gallery, Mail, Contacts, Calendar, and then the rest etc.

Please review @georgehrke @DeepDiver1975

### DIFF
--- a/appinfo/application.php
+++ b/appinfo/application.php
@@ -66,7 +66,7 @@ class Application extends App {
 
 		$server->getNavigationManager()->add(array(
 			'id' => $appName,
-			'order' => 10,
+			'order' => 5,
 			'href' => $server->getURLGenerator()
 				->linkToRoute('calendar.view.index'),
 			'icon' => $server->getURLGenerator()


georgehrke:
Is there some explanation about the order?

georgehrke:
:+1:

jancborchardt:
The explanation is:
- line 1: Files, Activity, Gallery – the apps which are shipped and enabled by default
- line 2: Mail, Contacts, Calendar – the typical personal info mgmt apps, most used (Contacts is before Calendar cause it relates strongly to Mail)
- line 3 etc: everything else (Notes, News, Maps, Music, Bookmarks etc)

cc @owncloud/designers @DeepDiver1975 for reference

jancborchardt:
Rebased, we can merge I guess.

skjnldsv:
Okay!
Are we planning on letting the user choose the order?

jancborchardt:
Nope, not for now at least and I actually don’t really see a reason to do that. (We had discussions about that in the past.) It’s just unnecessary work.
Maybe in the future we can introduce a »sort by most used« but that’s overkill too for now.

skjnldsv:
Agreed!

Issue #381, opened 2016-03-30
Sharing options overlap


nickvergessen:
### Steps to reproduce
1. Create a calendar
2. Type in a search string into the share field
### Expected behaviour

Beautiful list :heart_eyes: 
### Actual behaviour

![bildschirmfoto vom 2016-03-30 09 36 16](https://cloud.githubusercontent.com/assets/213943/14135027/2846aa4e-f65b-11e5-87ec-9ac533a1c72e.png)
### Server configuration

**Operating system**: Linux Mint 17.3

**Web server:** Apache2

**ownCloud version:** master 

**Calendar version:** master

**Updated from an older ownCloud or fresh install:** install

**List of activated apps:**

```
Enabled:
  - activity: 2.3.0
  - calendar: 1.0
  - comments: 0.3.0
  - dav: 0.2.2
  - federatedfilesharing: 0.2.0
  - files: 1.5.1
  - files_sharing: 0.10.0
  - files_texteditor: 2.1
  - files_trashbin: 0.9.0
  - files_versions: 1.3.0
  - provisioning_api: 0.5.0
  - systemtags: 0.3.0
  - updatenotification: 0.2.0
```
### Client configuration

**Browser:** Firefox 45.0, Chromium 49.0

**Operating system:** Linux Mint 17.3

georgehrke:
@raghunayyar

nickvergessen:
works, thanks :+1:

PR #412, opened 2016-04-03
Adds styles for sharing

raghunayyar:
Fine tunes the sharing drop-down styles.

Please check @georgehrke and @jancborchardt

### DIFF
--- a/css/calendarlist.css
+++ b/css/calendarlist.css
@@ -231,13 +231,20 @@ ul.dropdown-menu {
 	height: 100px;
     overflow: scroll;
     border: 1px #ddd solid;
-    border-radius: 3px;
+    border-radius: 0 0 3px 3px;
 	margin-top: -7px;
     margin-left: 6px;
-    width: 226px;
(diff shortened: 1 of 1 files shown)


georgehrke:
fixes https://github.com/owncloud/calendar/issues/381

georgehrke:
much better than before, but there are still some issues:
- [x] when typing a username the input field "jumps down" a few pixels until the results are loaded
- [x] There is some annoying whitespace in the bottom of the dialog, see:
  calendar_-_owncloud_chromium__today_at_10_56_57_am

raghunayyar:
Will fix this asap.

jancborchardt:
These styles should be in core so they are applied globally.

tomneedham:
> These styles should be in core so they are applied globally.

They would need to be adapted since contacts/calendar use an angular directive to perform the auto completion - but possible :)

georgehrke:
Let's fix the mentioned issues in a  follow up pr.

PR #413, opened 2016-04-03
bring back caldav link

### LINKED ISSUE #356: CalDAV Link for each Calendar missing
just another regression:

E.g. with Thunderbird SoGo Connector **we need the full CalDAV URL for each calendar**. The "primary CalDAV URL" is not usable with it.

Users must enter the full URL to get a working calendar with the sogo connector, which is a real PITA: You must read it from the calendar's tooltip, when hovering over the calendar name in the web app and then type it in by hand in the sogo connector configuration.

georgehrke:
bring back readonly input with caldav link due to popular demand
fixes https://github.com/owncloud/calendar/issues/356

### DIFF
--- a/css/globals.css
+++ b/css/globals.css
@@ -46,3 +46,13 @@
   background: url("../img/timezone.svg") center center no-repeat;
   display: block;
 }
+
+.input-with-button-on-right-side {
+  width: 85% !important;
+  float: left;
+}
+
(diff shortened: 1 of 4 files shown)


raghunayyar:
:+1:

brrrrrrrt:
:+1:   Thx!

PR #433, opened 2016-04-10
suggest different reminder options on default

### LINKED ISSUE #307: if a reminder with 15 mins before already exists, create one with 30 minutes before and so on

georgehrke:
fixes https://github.com/owncloud/calendar/issues/307


raghunayyar:
:+1:

PR #434, opened 2016-04-10
prevent accidental removal of events

### LINKED ISSUE #420: prevent accidental removal of events
When editing an existing event, the event is being deleted when I hit the enter key.
The default button should be "Update" in order to prevent accidental removals.

georgehrke:
fixes https://github.com/owncloud/calendar/issues/420


raghunayyar:
:+1:

PR #437, opened 2016-04-10
speed up event rendering by using renderEvent

### LINKED ISSUE #162: speed up event rendering by using renderEvent instead of returning array inside eventSource

georgehrke:
fixes https://github.com/owncloud/calendar/issues/162


georgehrke:
needs some more investigation, let's postpone this to 1.2

georgehrke:
I found a better solution, see webcal PR

PR #439, opened 2016-04-10
automatically add one hour when start and end equal

### LINKED ISSUE #216: Automatically add one hour to each event
As it is today you have to write in both the FROM time and the TO time table. 

It would be much better if you just had to write in the FROM time table and that one hour was automatically added. Often meetings aren't longer than 1 hour, and if you want to have a longer appointment, you simply edit the TO time manually.

georgehrke:
fixes https://github.com/owncloud/calendar/issues/216


raghunayyar:
:+1:

PR #443, opened 2016-04-11
Webcal support

### LINKED ISSUE #132: support for webcal
- [ ] add a subscription service
- [ ] for now query subscriptions on load (using render event with static option)

georgehrke:
fixes https://github.com/owncloud/calendar/issues/132

### DIFF
--- a/js/app/models/calendarmodel.js
+++ b/js/app/models/calendarmodel.js
@@ -50,6 +50,8 @@ app.factory('Calendar', ['$rootScope', '$filter', 'VEventService', 'TimezoneServ
 			tmpId: RandomStringService.generate(),
 			fcEventSource: {
 				events: function (start, end, timezone, callback) {
+					var fcAPI = this;
+
 					TimezoneService.get(timezone).then(function(tz) {
 						self.list.loading = true;
 						self.fcEventSource.isRendering = true;
@@ -62,6 +64,7 @@ app.factory('Calendar', ['$rootScope', '$filter', 'VEventService', 'TimezoneServ
 							}
 
 							callback(vevents);
+							fcAPI.reportEvents(fcAPI.clientEvents());
 							self.fcEventSource.isRendering = false;
 
 							self.list.loading = false;
--- a/js/app/models/subscriptionmodel.js
+++ b/js/app/models/subscriptionmodel.js
@@ -1,0 +1,80 @@
+/**
+ * ownCloud - Calendar App
+ *
+ * @author Raghu Nayyar
+ * @author Georg Ehrke
+ * @copyright 2016 Raghu Nayyar <beingminimal@gmail.com>
+ * @copyright 2016 Georg Ehrke <oc.list@georgehrke.com>
+ *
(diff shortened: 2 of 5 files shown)


georgehrke:
developer tools - http --owncloud dev-index php-apps-calendar- chromium today at 2 19 09 pm

To check: 
why is it d:collection and cal:calendar instead of the being replaced with the proper xml namespaces?

georgehrke:
why is it d:collection and cal:calendar instead of the short namespace being replaced with the long one?

developer_tools_-_http_--owncloud_dev-index_php-apps-calendar-_chromium__today_at_8_57_22_pm

@evert @PVince81 Is this the intended behavior of davclient.js or is it a bug?
(If it's a bug, I can look into it :wink:)

evert:
It's a bug!

PVince81:
Strange... it seems to work with the file list, see how it's checked here: https://github.com/owncloud/core/blob/v9.0.1/core/js/files/client.js#L286

Maybe the response contains newlines and the parsing fails for some reason

PVince81:
Or it's because there is more than one element, I never tested with more than one :frowning:

georgehrke:
While looking into fixing this I also noticed that `supported-calendar-component-set` was broken all the time. cc @DeepDiver1975 @raimund-schluessler 

developer_tools_-_http_--owncloud_dev-index_php-apps-calendar-_chromium__today_at_2_17_52_pm

georgehrke:
I would propose to replace these blocks with an array like:

```
{urn:ietf:params:xml:ns:caldav}supported-calendar-component-set: [
{
  propName: '{urn:item:params:xml:ns:caldav}comp',
  attributes: {
    name: 'VEVENT'
  }
},
{
  propName: '{urn:item:params:xml:ns:caldav}comp',
  attributes: {
    name: 'VTODO'
  }
}
],
```

Do we lose any data with this approach or is it fine?

georgehrke:
And especially when it comes to these nested acl blocks we need a proper recursive solution:

``` xml
    <d:acl>
     <d:ace>
      <d:principal>
       <d:href>/remote.php/dav/principals/users/admin/</d:href>
      </d:principal>
      <d:grant>
       <d:privilege>
        <d:read/>
       </d:privilege>
      </d:grant>
      <d:protected/>
     </d:ace>
     <d:ace>
      <d:principal>
       <d:href>/remote.php/dav/principals/users/admin/</d:href>
      </d:principal>
      <d:grant>
       <d:privilege>
        <d:write/>
       </d:privilege>
      </d:grant>
      <d:protected/>
     </d:ace>
    </d:acl>
```

PVince81:
@georgehrke mind having a look at the lib and submit a PR there if you find a fix that works ?

georgehrke:
Yes :)

just wanted to get feedback on my approach suggested in https://github.com/owncloud/calendar/pull/443#issuecomment-215065717

georgehrke:
closing in favor of https://github.com/owncloud/calendar/pull/631

PR #454, opened 2016-04-12
explicitly declare buttons as buttons to not make them trigger ng-submit

georgehrke:
don't trigger save action with add attendee or add reminder button


raghunayyar:
Yes please :+1:

georgehrke:
tests pass locally -> merge

Issue #453, opened 2016-04-12
use XMLSerializer instead of outerHTML prop


georgehrke:
see https://github.com/owncloud/calendar/issues/309#issuecomment-209060529

raghunayyar:
:+1:

PR #455, opened 2016-04-12
Update app.js

ansorg:
replace outerHTML on XML elements with XMLSerializer, to make it work in Edge
#453


georgehrke:
Sorry, but it won't work this way.
You need to update the calendarservice.js and build the app.js with `grunt build`. It's probably easier if I take care ;)

Thx anyway :)
With a bit luck we'll get this into 1.1 I was just about to package ;)

georgehrke:
Here we go: https://github.com/owncloud/calendar/pull/456

georgehrke:
@ansorg: otherwise your changes to app.js will be overwritten the next time I run `grunt build`

PR #456, opened 2016-04-12
be nice to edge users

### LINKED ISSUE #453: use XMLSerializer instead of outerHTML prop

georgehrke:
fixes https://github.com/owncloud/calendar/issues/453

### DIFF
--- a/js/app/service/calendarservice.js
+++ b/js/app/service/calendarservice.js
@@ -45,6 +45,8 @@ app.service('CalendarService', ['DavClient', 'Calendar', function(DavClient, Cal
 		'{' + DavClient.NS_OWNCLOUD + '}invite'
 	];
 
+	this._xmls = new XMLSerializer();
+
 	function discoverHome(callback) {
 		return DavClient.propFind(DavClient.buildUrl(OC.linkToRemoteBase('dav')), ['{' + DavClient.NS_DAV + '}current-user-principal'], 0, {'requesttoken': OC.requestToken}).then(function(response) {
 			if (!DavClient.wasRequestSuccessful(response.status)) {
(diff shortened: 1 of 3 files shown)


raghunayyar:
:+1: woot woot!

georgehrke:
tests pass locally -> merge

tigerfoot:
Even with this (in version 1.1.0) there an error on IE 11 (edge Mode)

SCRIPT5009: "XPathResult" ist undefiniert
Datei: client.js, Zeile: 278, Spalte: 9
var responseIterator = doc.evaluate('/d:multistatus/d:response', doc, resolver, XPathResult.ANY_TYPE, null);

Did we miss something ?

tigerfoot:
I've tried also to integrate wgxpath from https://github.com/google/wicked-good-xpath which could restore IE11 support with less effort.
but we a no 100% success (the name of the calendars appears, but not the event)
I'm using wgxpath.intall() at the beginning of app.js

In the faulty line in clients.js I was able to get a content. 

``` xml
xmlBody "<?xml version=\"1.0\"?>\n<d:multistatus xmlns:d=\"DAV:\" xmlns:s=\"http://sabredav.org/ns\" xmlns:cal=\"urn:ietf:params:xml:ns:caldav\" xmlns:cs=\"http://calendarserver.org/ns/\" xmlns:card=\"urn:ietf:params:xml:ns:carddav\" xmlns:oc=\"http://owncloud.org/ns\">\n <d:response>\n  <d:href>/remote.php/dav/</d:href>\n  <d:propstat>\n   <d:prop>\n    <d:current-user-principal>\n     <d:href>/remote.php/dav/principals/users/8886b774-e58d-102e-8af3-bd0838ca98dc/</d:href>\n    </d:current-user-principal>\n   </d:prop>\n   <d:status>HTTP/1.1 200 OK</d:status>\n  </d:propstat>\n </d:response>\n</d:multistatus>\n"
```

But the calendar doesn't still get displayed ...

PR #509, opened 2016-05-10
Adapt loader to owncloud/core#24491

skjnldsv:
Use the core classes instead of custom ones.
And prepare the calendar app to the future new loader (already merged in core/master)

Fully compatible with old AND new loading classes.

capture d ecran_2016-05-10_06-30-18

### DIFF
--- a/css/calendar.css
+++ b/css/calendar.css
@@ -22,10 +22,6 @@
  */
 
 
-#app-content.loading {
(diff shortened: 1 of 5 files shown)


georgehrke:
I like the new loaders a lot :)

Thx 👍

skjnldsv:
You're welcome :)

PR #513, opened 2016-05-11
update deps

georgehrke:
also explicitly add fullcalendar to use the latest version.


georgehrke:
@raghunayyar Mind reviewing?

PR #526, opened 2016-05-15
use random color as fallback when importing

### LINKED ISSUE #390: Imported calendars have same color
I just imported two calendars from .ics files and both have the same dark blue color.

For any new calendar or imported calendar, the color should automatically be set to the first unused color.
Also, we should not have this dark blue, still dark blue and the dark grey … they are much too dark and not distinguishable either. For example, we don’t have any green color at all.

georgehrke:
fixes https://github.com/owncloud/calendar/issues/390

### DIFF
--- a/js/app/controllers/importcontroller.js
+++ b/js/app/controllers/importcontroller.js
@@ -72,7 +72,7 @@ app.controller('ImportController', ['$scope', '$rootScope', '$filter', 'Calendar
 
 			if (file.calendar === 'new') {
 				var name = file.newCalendarName || file.name;
-				var color = file.newCalendarColor || '#1d2d44';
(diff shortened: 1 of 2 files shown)


aspdye:
Nice idea :+1:

raghunayyar:
XD

skjnldsv:
👍

georgehrke:
@skjnldsv randColour fails silently on ownCloud 9 and blocks everything 😕

skjnldsv:
Hum, any informations to add?
Why does it fail?

skjnldsv:
Is it because oc9 doesn't have the .toHsl function? ^^

Issue #531, opened 2016-05-17
show random color button for ownCloud 9.1+


georgehrke:
With 5bae6a108e2fb5937539d0d45bf8569bbec57b01 I hid the random color picker.
It doesn't work with ownCloud 9. We should only display it for ownCloud 9.1+

skjnldsv:
see #536

PR #536, opened 2016-05-18
Fix random colour if no toHsl function and import calendar loading

### LINKED ISSUE #390: Imported calendars have same color

skjnldsv:
- Generate random colour on import
- Test if core has the toHsl function
- If no core colour generator, random between our default list of colour
- Remove spinner once importing is finished

@georgehrke @jancborchardt @raghunayyar 
Fix #390 and #531
ref and replace #526


georgehrke:
works for me on stable9, thx :)

PR #547, opened 2016-05-19
remove ES6 array destructuring

### LINKED ISSUE #538: 1.2.1 calendar doesn't show up after upgrade
### Actual behaviour

Tell us what happens instead
The calendar app shows no calendars, and almost no interface.  everything is basically broken.
### Server configuration

**Operating system**:
debian 8
**Web server:**
apache
**Database:**
mysql
**PHP version:**
5.5
**ownCloud version:** (see ownCloud admin page)
9.0.2
**Calendar version:** (see ownCloud apps page)
1.2.1
**Updated from an older ownCloud or fresh install:**
updated

Is there an achieve of old version snapshots.  it would be nice to have if not so when an update breaks everything people can easily roll back.  Luckily I always back up apps before I upgrade them but it would still be nice just in case.

georgehrke:
fixes https://github.com/owncloud/calendar/issues/538

cc @vetienne @Luticus

### DIFF
--- a/js/app/directives/colorpickerDirective.js
+++ b/js/app/directives/colorpickerDirective.js
@@ -31,7 +31,9 @@ var hslToRgb = function(hue, saturation, lightness) {
 	'use strict';
 	// based on algorithm from http://en.wikipedia.org/wiki/HSL_and_HSV#Converting_to_RGB
 	if(Array.isArray(hue)) {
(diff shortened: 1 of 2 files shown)


georgehrke:
please review @skjnldsv @raghunayyar @jancborchardt

skjnldsv:
Perfect! 👍

PR #553, opened 2016-05-22
fixed typo "uknown"

michaelstingl:
'Editing reminders of unknown type not supported.'

Transifex user [corneliu.e](https://www.transifex.com/user/profile/corneliu.e/) found this [bug](https://www.transifex.com/owncloud-org/owncloud/translate/#ro/calendar/79316339).


georgehrke:
thx :)
👍

Issue #556, opened 2016-05-28
Calendars on the list are overlayed by subscription calendars


piesu:
### Steps to reproduce

1.Create 11 or more calendars
2.Have any non-modifiable calendar (birthdays, shared without edit permission)
3.Check calendar app
### Expected behaviour

You want to see all calenders correctly and be able to select every
### Actual behaviour

"Standard" calendars and "subscribed" overlay on each other
![problem screnshot](https://314es.pl/j/owncloud_calendars_problem.png)
### Server configuration

**Operating system**:
Ubuntu 16.04

**Web server:**
Apache

**Database:**
MySQL

**PHP version:**
PHP7

**ownCloud version:** 9.0.2

**Calendar version:** from deb, accorind to app:list version "true"?

**Updated from an older ownCloud or fresh install:**
updated (from owncloud 6? dunno)

**Signing status (ownCloud 9.0 and above):**
No errors have been found.

**List of activated apps:**
Enabled:
- activity: 2.2.1
- calendar: true
- comments: 0.2
- contacts: true
- dav: 0.1.6
- documents: true
- federatedfilesharing: 0.1.0
- federation: 0.0.4
- files: 1.4.4
- files_external: 0.5.2
- files_pdfviewer: 0.8.1
- files_sharing: 0.9.1
- files_texteditor: 2.1
- files_trashbin: 0.8.0
- files_versions: 1.2.0
- files_videoplayer: 0.9.8
- firstrunwizard: 1.1
- gallery: 14.5.0
- impersonate: 0.0.3
- news: true
- notifications: 0.2.3
- ownnote: 1.08
- provisioning_api: 0.4.1
- systemtags: 0.2
- templateeditor: 0.1
- updatenotification: 0.1.0
- user_external: 0.4
  Disabled:
- encryption
- external
- user_ldap

**The content of config/config.php:**
Would prefer not

**Are you using external storage, if yes which one:** local/smb/sftp/...
no

**Are you using encryption:** yes/no
no

**Are you using an external user-backend, if yes which one:** LDAP/ActiveDirectory/Webdav/...
no
### Client configuration

**Browser:**
Google Chrome 50

**Operating system:**
Debian testing

**CalDAV-clients:**
Owncloud web interface
#### Web server error log

it's css issue
#### ownCloud log (data/owncloud.log)

it's css issue
#### Browser log

it's css issue

georgehrke:
dup of https://github.com/owncloud/calendar/issues/552

[...]

Issue #552, opened 2016-05-21
Left navigation bar styling issue


onny:
### Steps to reproduce
1. Open the calendar app
2. Create as many calendars until the vertical space of the left navigation is exceeded
3. Now the last entries on the bottom starts overlapping
### Expected behaviour

Left navigation menu should be scrollable in case there are too many entries or the screen size is too small
### Actual behaviour

There seems to be some styling issues in the sidebar ...
![2016-05-21-125934_swaygrab](https://cloud.githubusercontent.com/assets/757752/15448167/7022fb6a-1f5a-11e6-9dcc-c5532be608d4.png)
### Server configuration

**Operating system**: ArchLinux x74

**Web server:** nginx 1.10.0

**Database:** mysql-server 5.5.47

**PHP version:** php-fpm 7.0.6

**ownCloud version:** 9.0.2

**Calendar version:** 1.2.2
**Updated from an older ownCloud or fresh install:** Updated from older Owncloud version
### Client configuration

**Browser:** firefox 46.0.1
**Operating system:** ArchLinux x64

georgehrke:
@raghunayyar please take a look :)

onny:
Looks like this bug is fixed with version 1.2.2. Thanks a lot!

georgehrke:
Okay 🤔

PR #557, opened 2016-05-28
Remove unnacesary max-height for calendars list

piesu:
Fix for https://github.com/owncloud/calendar/issues/552 and https://github.com/owncloud/calendar/issues/556


georgehrke:
thx 👍

PR #566, opened 2016-06-05
Add an indication to the calendar list when a calendar is shared

MathieuSchopfer:
This PR is related to issue #431.

PR #571, opened 2016-06-09
Add an indication to the calendar list when a calendar is shared

### LINKED ISSUE #431: add proper indication if a calendar was shared
change the opacity to something higher and add `Shared` label

georgehrke:
fixes https://github.com/owncloud/calendar/issues/431
supersedes https://github.com/owncloud/calendar/pull/566

### DIFF
--- a/css/calendarlist.css
+++ b/css/calendarlist.css
@@ -112,6 +112,9 @@
 #app-navigation .app-navigation-list-item .utils .action span {
 	opacity: .3;
 }
+#app-navigation .app-navigation-list-item .utils .action.withitems span {
+	opacity: .7 !important;
+}
 #app-navigation .app-navigation-list-item .utils .action:hover span {
 	opacity: 1;
 }
--- a/templates/part.calendarlist.item.php
+++ b/templates/part.calendarlist.item.php
@@ -33,13 +33,21 @@
 	{{ calendar.displayname }}
 </a>
 <span class="utils" ng-if="!calendar.list.locked" ng-show="!calendar.list.edit">
-	<span class="action">
+	<span class="action" ng-class="{'withitems': calendar.hasShares()}">
 		<span
 			ng-if="calendar.shareable"
-			class="calendarlist-icon share icon-share permanent"
+	 		class="calendarlist-icon share permanent"
+			ng-class="{'icon-shared': calendar.hasShares(), 'icon-share': !hasShares(calendar)}"
 			data-item-type="calendar" data-item="{{ calendar.id }}"
 			title="<?php p($l->t('Share Calendar')) ?>"
 			ng-click="toggleSharesEditor(calendar)">
+		</span>
+		<!-- Add a label if the calendar has shares -->
+		<span
+			ng-if="calendar.hasShares() && calendar.shareable"
+			style="position: relative; bottom: 16px; left: -12px; width: 30px; opacity: 1"
+			ng-click="toggleSharesEditor(calendar)">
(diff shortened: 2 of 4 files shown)


templates/part.calendarlist.item.php
```diff
@@ -33,14 +33,22 @@ class="icon icon-error">
...
+	 		class="calendarlist-icon share permanent"
+			ng-class="{'icon-shared': calendar.hasShares(), 'icon-share': !hasShares(calendar)}"
 			data-item-type="calendar" data-item="{{ calendar.id }}"
 			title="<?php p($l->t('Share Calendar')) ?>"
 			ng-click="toggleSharesEditor(calendar)">
 		</span>
+		<!-- Add a label if the calendar has shares -->
+		<span
+			ng-if="calendar.hasShares() && calendar.shareable"
+			style="position: relative; bottom: 16px; left: -12px; width: 30px; opacity: 1"
```
georgehrke, line 48: @raghunayyar @jancborchardt Do you have a somewhat more elegant solution for this? :)

css/calendarlist.css
```diff
@@ -112,6 +112,9 @@
 #app-navigation .app-navigation-list-item .utils .action span {
 	opacity: .3;
 }
+#app-navigation .app-navigation-list-item .utils .action.withitems span {
+	opacity: .7 !important;
```
MathieuSchopfer, line 116: The `!important` rule overrides the `:hover` selector below, so the icon opacity won't turn to 1 when the mouse hovers over it.

georgehrke, line 116: Yes, just like in the files app :) 

Please excuse my brevity and typos.

MathieuSchopfer:
Okay ;)

css/calendarlist.css
MathieuSchopfer: Okay ;)

templates/part.calendarlist.item.php
georgehrke, line 48: ping @jancborchardt

jancborchardt:
For any pull request with interface changes, please ideally include a screenshot. :) That makes it much easier to see what changed and quicker to review.

Generally speaking, the indicator should be like in the Files app: The share icon being permanently displayed.

MathieuSchopfer:
@jancborchardt Here you go:
shared_cal

> Generally speaking, the indicator should be like in the Files app: The share icon being permanently displayed.

Yes, this is what we did :)
The problem concerns the position of the _Shared_ label, which required some extra (ugly) styling :/

templates/part.calendarlist.item.php
raimund-schluessler, line 48: Maybe you could consider using `ng-class` to not use inline-css? And use `display: none` to not show it when not necessary.

georgehrke, line 48: `ng-class` is not necessary here, a `class` would do the same, but it's rather about the style itself.
There are probably better solutions than hardcoding some pixel values 😉

jancborchardt:
@MathieuSchopfer perfect! :) Good stuff and :+1: from the design side.

georgehrke:
@jancborchardt Do you have a better solution for `style="position: relative; bottom: 16px; left: -12px; width: 30px; opacity: 1"`?
It's not the most elegant solution 😉

jancborchardt:
Lets go with that for now, we can always improve later.

georgehrke:
okay :)
looking into the conflicts ...

jancborchardt:
Awesome work everyone :)

PR #579, opened 2016-06-15
Display description in popover

raoulbhatia:


raoulbhatia:
I am more heavily using the description than the location of my appointments to add additional information that I do not want to clutter the event's title with.

templates/editor.popover.php
```diff
@@ -35,6 +35,8 @@ class="events--input h2"
 		</fieldset>
 
 		<fieldset class="events--fieldset" ng-disabled="readOnly">
+			<textarea ng-model="properties.description.value" type="text" class="advanced--input advanced--textarea"
```
georgehrke: `advanced--input` and `advanced--textarea` are for the sidebar only and I don't think they have any effect here. Take a look at the location input for the appropriate classes.

georgehrke:
@raghunayyar @jancborchardt Do we want the description in the popover?

raoulbhatia:
Thanks for the review - I have amended my commit.

georgehrke:
Thx for your contribution! :)
Will test asap.

georgehrke:
@raoulbhatia Please put the description field below the location field. (Just like it's in the sidebar 😉)
Having different orders in popover and sidebar would be confusing.

This increases the size of the popover a lot though. 😕 
combined

ping @jancborchardt @raghunayyar

jancborchardt:
Sorry but the description field has no place in the **quick add/edit** popover. That popover needs to be as simple and concise as possible, so I would actually say we go even simpler by removing the »Location« field as well: https://github.com/owncloud/calendar/issues/588

raoulbhatia:
How about making this configurable?
I.e. that one can choose what to display by default in the pop over?

Thanks,
Raoul

jancborchardt:
We aim for the apps to be easily usable and streamlined. That also means to not overload the settings with options which are not really necessary. Because then you won’t find the really important options anymore. If you really want it, you can still simply apply should short patch. :)

(Sorry if this sounds negative, it’s not meant like that and we value contributions a lot! :) However we have to make sure we stay clear to our core mission to deliver a very easy to use software.)

raoulbhatia:
So this means that it is rejected, right? :-(

georgehrke:
I have to agree with jan that we don't want to clutter the popover with more inputs. 

I would propose one tradeoff though. Instead of making the popover bigger just add an option to skip the popover in its entirety and always open the sidebar.

@jancborchardt What do you think about this? ^
We also had feature requests for this if i recall correctly.

raoulbhatia:
@georgehrke this would perfectly work for me and would be awesome as an option!

jancborchardt:
@georgehrke maybe that makes sense, yeah. The default should be the popover though of course.

georgehrke:
> The default should be the popover though of course.

Sure :)

georgehrke:
I created a dedicated issue: https://github.com/owncloud/calendar/issues/597

Issue #485, opened 2016-04-21
Improvement: parse time in event title prompt


mcorteel:
Hello!

Great work on the new calendar, it's awesome. One time-saving feature I really like in Google calendar is the ability to set the time for an event on creation by typing it before the event title.
For example "14:00 Dentist" will be an event that starts at 14:00, with a default duration (for example 1 hour) with the title "Dentist".

What do you think?

georgehrke:
Hm, the same feature is also in OS X calendar and Fantastical, but I never use it there. In Fantastical it actually annoys me a bit, because it's in the title. (whereas OS X calendar has a dedicated button / input for it).

This is also very difficult, because we would have to adapt different locals. Users will expect it to work with their locale time format.

Also selecting a start time is three clicks at most, so it's not that time-saving.

If anything this would be assigned a very very low priority for now, but I'll let @jancborchardt decide if we want this feature.

@jancborchardt ^

jancborchardt:
Yeah, I would also say it’s low-prio since it’s _a lot_ of work. And when we implement it, people will expect to work it in all kinds of different scenarios and writing styles, otherwise it’s not proper. And since it’s such a fringe feature, there’s a ton of other things which are more important.

@mcorteel contributions are always welcome though of course! :) So if you want to get your feet wet in Calendar development, maybe this is an interesting thing? I suggest to check out some starter issues first though: https://github.com/owncloud/calendar/issues?q=is%3Aissue+is%3Aopen+label%3A%22starter+issue%22

mcorteel:
I am interested and I have considered helping with the development of such a feature. Maybe someday!
I actually met some of you in Berlin two years ago, but I didn't have the time or knowledge to help much. I'm getting there though :)
@georgehrke the number of clicks is exactly my problem. Moving from keyboard to mouse is tiring for a lazy guy like me... But I know that most users don't care!

jancborchardt:
@mcorteel if you are still in Berlin, check out the monthly Berlin meetup: http://meetup.com/owncloud-berlin
Or the upcoming contributor conference! https://owncloud.org/conf

mcorteel:
It would have been great, but I live in Ludwigshafen now!

yeKcim:
"14:00 Dentist" with default hour setting is great
"14:00-16:00 Dentist" is really great

This feature is why I continue to use google calendar :'-(

jancborchardt:
[...]

PR #581, opened 2016-06-16
Parse time schedule in the event summary

MathieuSchopfer:
Implements feature request #485.
See this [fiddle](https://jsfiddle.net/MathieuSchopfer/0co07f8p/) for an equivalent example.

### DIFF
--- a/js/app/controllers/editorcontroller.js
+++ b/js/app/controllers/editorcontroller.js
@@ -85,6 +85,44 @@ app.controller('EditorController', ['$scope', 'TimezoneService', 'AutoCompletion
 		$scope.registerPostHook = function(callback) {
 			$scope.postEditingHooks.push(callback);
 		};
+		
+		$scope.parse_summary = function() {
+			var summary = $scope.properties.summary.value;
+			
+			// Parse text to extract schedule from string beginning
+			var timePattern = "(\\d{1,2}:\\d{2}(?:\\s?[ap]\\.?m\\.?)?)";
+			var linkPattern = "\\s?[-/]\\s?";
+			var scheduleExp = new RegExp("^"+timePattern+"(?:"+linkPattern+timePattern+")?\\s", 'i');
+			var matches = scheduleExp.exec(summary);
+			
+			if(matches !== null) {
+		
+			  	// Strip schedule string from summary
+			  	summary = summary.substring(matches.shift().length);
+			
+			  	var start = matches[0];
+			  	var end = matches[1];
+				
+				// Convert string to time
+				var schemes = ["HH:mm", "hh:mm a"];
+				start = moment(start, schemes);
+				if(typeof end !== 'undefined') {
+					end = moment(end, schemes);
+				}
+				else {
+					end = start.clone();
+				}
+				
+				// Update event properties
+				$scope.properties.dtstart.value = moment($scope.properties.dtstart.value.set(
+						{'hour': start.get('hour'), 'minute': start.get('minute')}));
+				$scope.properties.dtend.value = moment($scope.properties.dtend.value.set(
+						{'hour': end.get('hour'), 'minute': end.get('minute')}));
+				$scope.properties.allDay = false;
+				$scope.toggledAllDay();
+				$scope.properties.summary.value = summary;
+			}
+		};
 
 		$scope.proceed = function() {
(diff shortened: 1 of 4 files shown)


js/app/controllers/editorcontroller.js
```diff
@@ -79,6 +79,43 @@ app.controller('EditorController', ['$scope', 'TimezoneService', 'AutoCompletion
...
+				start = moment(start, schemes);
+				if(typeof end !== 'undefined') {
+					end = moment(end, schemes);
+				}
+				else {
+					end = start.clone().add(1, 'hour');
+				}
+				
+				// Update event properties
+				$scope.properties.dtstart.value = moment($scope.properties.dtstart.value.set(
```
MathieuSchopfer: I would have preferred to directly update the objects `$scope.properties.dtstart/dtend.value` using their [set](http://momentjs.com/docs/#/get-set/set/) method rather creating new ones, but this wouldn't work. I don't understand why ...

js/app/controllers/editorcontroller.js
```diff
@@ -79,6 +79,44 @@ app.controller('EditorController', ['$scope', 'TimezoneService', 'AutoCompletion
...
+				start = moment(start, schemes);
+				if(typeof end !== 'undefined') {
+					end = moment(end, schemes);
+				}
+				else {
+					end = start.clone();
+				}
+				
+				// Update event properties
+				$scope.properties.dtstart.value = moment($scope.properties.dtstart.value.set(
```
MathieuSchopfer, line 117: I would have preferred to directly update the objects `$scope.properties.dtstart/dtend.value` using their [set](http://momentjs.com/docs/#/get-set/set/) method rather creating new ones, but this wouldn't work. I don't understand why ...

js/app/controllers/editorcontroller.js
```diff
@@ -79,6 +79,44 @@ app.controller('EditorController', ['$scope', 'TimezoneService', 'AutoCompletion
...
+						{'hour': start.get('hour'), 'minute': start.get('minute')}));
+				$scope.properties.dtend.value = moment($scope.properties.dtend.value.set(
+						{'hour': end.get('hour'), 'minute': end.get('minute')}));
+				$scope.properties.allDay = false;
+				$scope.toggledAllDay();
```
MathieuSchopfer, line 122: Call to `$scope.toggledAllDay()` is useful to ensure that the default time interval (1 hour) is defined in just one place.

georgehrke:
cc @jancborchardt

jancborchardt:
Woah nice @MathieuSchopfer! :)

cc @mcorteel @yeKcim please test this as you requested this in #485 

(@MathieuSchopfer if you fix an issue ideally always @-mention the people who were involved in the issue so they see the fix. :)

jancborchardt:
From my side looking at the fiddle it seems good :+1: 
I guess this is mostly something we have to try out in normal use if it works fine.

georgehrke:
will test this tomorrow

raghunayyar:
Please rebase

MathieuSchopfer:
Done ;-)

raghunayyar:
@georgehrke can we have another pair of eyes here? :)

One sec, the diff seems weird though with the public/app.js compilation.

georgehrke:
please remove the app.js 
(it was deleted in master)

georgehrke:
Hm, what if someone want's to have a title like "15:00 title".
It'll always get replaced. :/

raimund-schluessler:
Maybe introduce a setting to turn off automatic parsing? :see_no_evil:

georgehrke:
I would prefer a dropdown that shows different options what to do.
There is an issue about that. Mostly related to entering titles like dinner and automatically suggesting setting the time to 8pm or something similar

raimund-schluessler:
I haven't had time to look at this PR, but I guees since `ng-blur` is used it only updates when the input field loses focus.

Hence, one improvement may be to evaluate the summary on keypress to directly show what the user can expect.
Also, one could think about only parsing the summary in the popover editor but not in the sidebar as a help to quickly create an event. Thus it would still be possible to change the title e.g. to `15:00 title` in the sidebar later on (this is similar to how e.g. Fantastical 2 does it on iOS).

PR #586, opened 2016-06-23
Fix typo

tcitworld:
Just a small error.


georgehrke:
good catch! 
thx for fixing it :)

georgehrke:
travis error unrelated, merging.

Thx for your contribution!

PR #593, opened 2016-06-23
use proper release of ical.js

### LINKED ISSUE #539: use latest ical release instead of master

georgehrke:
fixes https://github.com/owncloud/calendar/issues/539


georgehrke:
failing test is unrelated

PR #595, opened 2016-06-23
exclude appinfo from code coverage

georgehrke:


georgehrke:
- [x] Can we simply add this to phpunit.xml?

PR #602, opened 2016-07-03
fix dynamic colorpicker layout

EtlamGit:
This allows random number of colors for the colorpicker, without destroying the dynamically created layout.

Before this commit a change to calendar colors can destroy the layout of the calendar list. It may overlap with the calendar to the right and everything gets unreadable.
Example BEFORE:
before

With this change the colorpicker colors are arranged automatically to fit the desired space. The original layout is preserved completely.
Example AFTER:
after

Change in detail:
- The layout gets cluttered because the colors are not wrapping, but appended to the right infinitely. Therefore the `flex-wrap: wrap;` is necessary.
- To get the height to adapt automatically we need `height: auto;`
- And to preserve the original spacing we have to add some padding. It was 30px - 24px (for the color), therefore we need 3px at top and bottom.


georgehrke:
please review @jancborchardt

raghunayyar:
Please rebase @EtlamGit

georgehrke:
@raghunayyar @jancborchardt please review :)

The commit history is a bit fucked up though, but that's not a big deal.
There is only one commit: https://github.com/owncloud/calendar/pull/602/commits/ce617b8a688e23a4509ee38eab0e39a589acdff0
which can easily be extracted to be merged into master.

EtlamGit:
Yes, rebase always produced unresolvable errors. Don't know what happened there, as it is working in other projects.

If there is some indication that that single commit may be accepted, I could also redo that simple change on a clean fork and with a new pull request.

raimund-schluessler:
:+1: 

Not really necessary at the moment, but it does improve the layout very much if more colors are present.

georgehrke:
okay, I'll merge this manually :)

georgehrke:
https://github.com/owncloud/calendar/commit/8579c3e052897e253a4079372a3f15df5b9289f1 and  https://github.com/owncloud/calendar/commit/bc9a93b9b3e7d2b7d51189c66ec8a175855d3671

PR #605, opened 2016-07-04
Calendar public sharing

tcitworld:
PR associated with https://github.com/owncloud/core/pull/25351

Note that this PR can be merged before the one in core (it doesn't bring errors).
## Brings

a checkbox to set a calendar public, activated by the Share button, above the share input.
demo1

When the checkbox is checked, you have this message and a link to the public page to view your calendar
demo2

Public (not authenticated access) full view looks like this. The header still needs to be changed (it has the app buttons on the left, and the help and disconnect items inside the right menu). **Fixed**.
demo3

Left panel looks like this. We have the CalDAV uri access for clients (which sometimes need dummy credentials, though), the link to download the file as .ics (right now it's filename is the token) and a integration code. 
demo4

Here, we have an event, with it's details (as read-only of course).
demo5

TODOs : 
- [x] Fix header on public page
- [x] Fix double slashes inside url
- [x] public ics filenames : Needs to be done on core's side by implementing the ICSExportPlugin
- [x] share buttons (like in https://github.com/owncloud/core/pull/25576 - maybe wait for it to be merge, maybe we can reuse icons) _Will add after_
- [x] translations (are they done automatically ?)


tcitworld:
It's finished for now. Only issue, a publication is shown also as a share with the user itself. This is because I use inside https://github.com/owncloud/core/pull/25351 the `oc_dav_share` table to store the publications, and this may change.

georgehrke:
Where is the big difference between `PublicController` and `CalController` respectively `PublicInformationsController` and `CalendarListController`?

We really have lots of duplicate code there :/

tcitworld:
Difference is essentially that CalendarService needs to be called on a specials url `/dav/public-calendars/<token>` and mustn't PROPFIND on `/dav`. I don't know how to pass the information to CalController that we're on a public page and it must do things differently.

georgehrke:
@tcitworld Do you mind if I commit on this branch directly (tomorrow)?
I would like to get rid of the duplicate code. :)

tcitworld:
Not at all !

tcitworld:
Had a chance to take a look @georgehrke ?

georgehrke:
✅ Publish checkbox not visible on master core

georgehrke:
One issue though:
The icon next to the shared label should change to the chain when published. 
So it has exactly the same behavior like in the files app.

tcitworld:
Should be good.

js/app/controllers/calcontroller.js
```diff
@@ -130,19 +130,37 @@ app.controller('CalController', ['$scope', '$rootScope', '$window', 'Calendar',
...
-			$scope.calendars = calendars;
-			is.loading = false;
-			// TODO - scope.apply should not be necessary here
-			$scope.$apply();
-
-			angular.forEach($scope.calendars, function (calendar) {
-				$scope.eventSource[calendar.url] = calendar.fcEventSource;
-				if (calendar.enabled) {
+		var url = window.location.toString();
+		if (url.substr(url.length - 9) === 'calendar/' || url.substr(url.length - 8) === 'calendar' || url.substr(url.length - 10) === 'calendar/#') {
```
tcitworld: There must be another way to do this.

georgehrke: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/endsWith ? 😉

tcitworld: Oh yeah, it didn't used it at first because of compatibility, but now that we have Babel...

georgehrke: We don't support IE anyway ¯_(ツ)_/¯

tcitworld: Yup, but Android, Safari Mobile, Opera...

georgehrke: mobile Safari supports it according to http://kangax.github.io/compat-table/es6/#test-String.prototype_methods_String.prototype.endsWith

I don't use Android, how common is it to use the stock Android browser?

tcitworld: Indeed, I just looked on the MDN.

Let's just say people who use it are not exactly the ones using ownCloud. :-)

tcitworld:
Rebased and squashed.

tcitworld:
@georgehrke  When a calendar is published, I want to send some people an email saying the calendar has been published. I added a button next to the link of the published calendar like this.
mailing-btn

Now, should this button toggle the display of another text input field (just like it's done in the File app) or should it open a modal ?
I fear that with the first solution, users would get lost with the regular sharing field right below. In that case, maybe there should be some kind of separator between the publishing and the sharing part.

Right now I'm using a `prompt()` to ask for recipient's email. Please have a look at the public-send-emails branch to see how I'm doing it on php's side.

js/app/controllers/calcontroller.js
```diff
@@ -254,7 +273,9 @@ app.controller('CalController', ['$scope', '$rootScope', '$window', 'Calendar',
 					});
 
 					if (writableCalendars.length === 0) {
-						OC.Notification.showTemporary(t('calendar', 'Please create a calendar first.'));
+						if ($scope.calendars[0].publicurl !== $window.location.href) {
```
georgehrke: What is this supposed to check?

js/app/service/timezoneservice.js
```diff
@@ -48,7 +48,7 @@ app.service('TimezoneService', ['$rootScope', '$http', 'Timezone', 'TimezoneList
 
 			_this._timezones[tzid] = $http({
 				method: 'GET',
-				url: $rootScope.baseUrl + 'timezones/' + tzid + '.ics'
+				url: OC.generateUrl('/apps/calendar/v1/timezones/') + tzid + '.ics'
```
georgehrke: this won't work when calendar is in a different app folder
Let's fix `$rootScope.baseUrl` instead ;)

georgehrke:
@tcitworld Maybe take a look at https://github.com/owncloud/core/pull/25576
@DeepDiver1975 is just using `mailto:` there: https://github.com/owncloud/core/pull/25576/files#diff-50fd767aacbe8001294d17e5e21efbf8R33

georgehrke:

georgehrke:
calendar_-_owncloud_chromium__today_at_7_17_50_pm
- [x] please remove the button style (grey background)

georgehrke:
owncloud chromium today at 7 19 55 pm
- [x] please add some space on the left side of the labels

georgehrke:
- [x] Rename `Export to ICS` to `Download`? cc @jancborchardt

tcitworld:
> @tcitworld Maybe take a look at owncloud/core#25576
> @DeepDiver1975 is just using mailto: there: https://github.com/owncloud/core/pull/25576/files#diff-50fd767aacbe8001294d17e5e21efbf8R33

I wanted two things : 
- Send emails through ownCloud like the Files app does if the _Allow users to send mail notification for shared files_ admin option is activated.
- Provide share buttons (including a _mailto:_ link) on the public page like on owncloud/core#25576 (maybe reuse them).

js/app/controllers/calcontroller.js
tcitworld: It was to hide the notification if we are on a public page. Will change.

tcitworld:
- [x] Events on public page are currently editable.

tcitworld:
Rebased.

georgehrke:
Maybe we should display the webcal address as well, so you can add it to other ownCloud's once the webcal feature is ready

tcitworld:
Webcal address of the .ics file you mean ? In an input field too ?

Maybe we should have a separate discussion about this and federation (a « add to my ownCloud button such the files app has ?).

georgehrke:
> Maybe we should have a separate discussion about this and federation (a « add to my ownCloud button such the files app has ?).

Definitely. Federated calendar and contacts sharing is something we want for the future, but it requires to implement a caldav and carddav **client** in the dav app.
That's probably not something that will make it into 9.2.

cc @DeepDiver1975 ^

tcitworld:
Here's what it looks like.
sidebar1
sidebar2

georgehrke:
> translations (are they done automatically ?)

As long as you use `p($l->t('...'));` in PHP and `t()` or `n()` in JS the strings will automatically be published on Transifex and translated by the community.

tcitworld:
Thanks, that's what I thought.

tcitworld:
Send emails to tell about the newly published calendar.
emailpublic

georgehrke:
please review @raghunayyar @jancborchardt @raimund-schluessler @DeepDiver1975 :)

georgehrke:
- [x] fix $rootScope.baseUrl

georgehrke:
calendar - owncloud chromium today at 6 30 48 pm
the send button is misaligned in the latest chromium

tcitworld:
I made a dirty hack to fix it with Firefox (position relative, top), and it results in this. Will change.

raimund-schluessler:
Seems to work nicely in general. A few things I noticed:
- What happens if a calendar is unpublished? I can access basically any URL under `owncloud/index.php/apps/calendar/public/xxxxxxxxxxx` without any notice that there is no calendar accessable (the CalDAV url and the iframe code are not shown, but this might be overlooked). For a user this might look as if there are not entries in the respective calendar at all. Maybe some kind of notice or not showing the calendar at all would be helpful.
- It might be reasonable to hide the iframe-code and CalDAV link in the settings view as it is done in the normal private calendar view. What does @jancborchardt think?
- Is it possible to show the name of the shared calendar anywhere?
- Also I don't see the send email stuff at the moment. I will have a look the next days, why it is not shown.

georgehrke:
rebased

raghunayyar:
Please rebase again. :)

tcitworld:
> What happens if a calendar is unpublished? I can access basically any URL under owncloud/index.php/apps/calendar/public/xxxxxxxxxxx without any notice that there is no calendar accessable (the CalDAV url and the iframe code are not shown, but this might be overlooked). For a user this might look as if there are not entries in the respective calendar at all. Maybe some kind of notice or not showing the calendar at all would be helpful.
> 
> Is it possible to show the name of the shared calendar anywhere?

:+1: Will do.

> Also I don't see the send email stuff at the moment. I will have a look the next days, why it is not shown.

Don't forget to build js files. :wink:

jancborchardt:
> It might be reasonable to hide the iframe-code and CalDAV link in the settings view as it is done in the normal private calendar view. What does @jancborchardt think?

Good call @raimund-schluessler! @tcitworld can you do that? :)

tcitworld:
Rebased and handle the can-be-published property (so you'll need to update the core branch).

> > It might be reasonable to hide the iframe-code and CalDAV link in the settings view as it is done in the normal private calendar view. What does @jancborchardt think?
> 
> Good call @raimund-schluessler! @tcitworld can you do that? :)

Sure, but it leaves only the calendar title and the download button in the sidebar, and should be quite empty.

tcitworld:
Now that `CS:allowed-sharing-modes > can-be-shared` is provided, should the `props.shareable` use this property instead of relying on write access ?

georgehrke:
Yes, that would be the best :)

tcitworld:
However, that would make the calendar app not working until the core PR is merged.

georgehrke:
Can you please add a fallback, that sets shareable to `publicMode ? false : props.canWrite` when `allowed-sharing-modes` is not set?
We need this, because the master branch needs to continue working on ownCloud 9.0 and 9.1

tcitworld:
> We need this, because the master branch needs to continue working on ownCloud 9.0 and 9.1

Also, the _Contacts Birthdays_ calendar is produced as a regular calendar from _core_ but has to be shown as a subscription in the app.

tcitworld:
@jancborchardt As I said, don't really know how to fill the empty space here...
cal-public-view-w-settings-1
cal-public-view-w-settings-2

georgehrke:
> Also, the Contacts Birthdays calendar is produced as a regular calendar from core but has to be shown as a subscription in the app.

Subscriptions are just read-only calendars. There is no direct connection to webcal calendars. Read-only shares also appear in subscriptions.

tcitworld:
Yes, I was just saying read-only calendars should not (and server can't anyway) be shared or published.

raghunayyar:
Do you guys need me here?

tcitworld:
@raghunayyar  If you have UI ideas to fill the left sidebar. :-)

tcitworld:
Rebased.

tcitworld:
Why does Coverall gets crazy like that ?

tcitworld:
I want to move the mail content into a special template file (so it would be easier to change for admins who want to customize them). Any idea on how to do that apart from `eval()` an html file ?

georgehrke:
You can include other template files with `<?php inc('filename'); ?>`

tcitworld:
Directly from the controller ? I didn't even think to test, but it's great if I can do if from here.

Edit : TemplateResponse->render() should be what I'm needing.

tcitworld:
Looks quite nice.
screenshot_20160817-190337

jancborchardt:
@tcitworld nice! To avoid the he/she stuff, let’s just shorten the text to: »username shared the calendar **test** with you.«

tcitworld:
> @tcitworld nice! To avoid the he/she stuff, let’s just shorten the text to: »username shared the calendar test with you.«

@jancborchardt That's what I had put first : but that's not really the case - the calendar is public - so _username_ shared the calendar with everyone.

EDIT : I'm going with `We wanted to inform you that %user% has publicly shared the calendar %calendar%`

georgehrke:
We can replace the he/she with they.

I'm not a native speaker, but I think we should put the 'publicly' at the end of the sentence.

georgehrke:
cal-public-view-w-settings-1

I think the text is way to big.
Just use the same font size as in the normal view :)

cc @jancborchardt

georgehrke:
- [ ] And there needs to be some error handling, when the shared calendar doesn't exist.
  e.g. when it was unshared :)

tcitworld:
> when the shared calendar doesn't exist. e.g. when it was unshared :)

I can't differentiate those two, but you're right, I had forgotten about them.

georgehrke:
> I can't differentiate those two

Yes, but there has to be some feedback for the user that this shared calendar doesn't exist :)

georgehrke:
Maybe we should change the naming from [ ] publish to [ ] Share link and put it under the internal sharing foo, so it's consistent with the files app. cc @jancborchardt 

files - owncloud safari technology preview today at 12 22 37 pm

tcitworld:
> And there needs to be some error handling, when the shared calendar doesn't exist. e.g. when it was unshared :)

How would this be shown ? A prompt message box in the middle of the window ? @georgehrke @jancborchardt

ulimoebius:
is there a release date or a version-number when this feature shown up in the stable release?

georgehrke:
This requires server changes -> ownCloud 9.2 / Nextcloud 11

tcitworld:
- [x] Core drops `pre-publish-url` support so we can remove it here https://github.com/owncloud/core/pull/26023

tcitworld:
Since we don't anymore know where the calendar is being published, we need to call PROPFIND again to see where it has been published.

tcitworld:
Needs core fixes, but ready on this side : https://github.com/owncloud/core/pull/26023/files/416306fd72dc0d62342e88e8e9b999b1fdf8dd20#r77823913

tcitworld:
> And there needs to be some error handling, when the shared calendar doesn't exist. e.g. when it was unshared :)

Seems like `OC.Notification` notifications don't show up on a public template.

georgehrke:
Is this ready from your point of view?

tcitworld:
- [x] Rebase
- [x] Fix tests
- [ ] Show a notification if no calendar exists at this url (with bootstrap components, since you can't display a notification through oC)
- [ ] Test when core fix is here

tcitworld:
@georgehrke Well, we should be not far from good.

tcitworld:
Closing, will come upstream.

PR #614, opened 2016-07-11
add XMLUtility

### LINKED ISSUE #590: create XMLUtility to convert json to xml
example parameters:

``` javascript
foo({
name:
attributes: [],
children: []
});
```

georgehrke:
fixes https://github.com/owncloud/calendar/issues/590

let's get rid of all this XML 💩 in the services

### DIFF
--- a/js/app/utility/xmlUtility.js
+++ b/js/app/utility/xmlUtility.js
@@ -1,0 +1,92 @@
+/**
+ * ownCloud - Calendar App
+ *
+ * @author Georg Ehrke
+ * @copyright 2016 Georg Ehrke <oc.list@georgehrke.com>
+ *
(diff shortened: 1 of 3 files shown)


georgehrke:
ToDo:
- [x] add `getRootSceleton('d:mkcol', 'd:set', 'd:prop');` that will return:

```
{
            name: 'd:mkcol',
            attributes: {
                'xmlns:c': 'urn:ietf:params:xml:ns:caldav',
                'xmlns:d': 'DAV:',
                'xmlns:a': 'http://apple.com/ns/ical/',
                'xmlns:o': 'http://owncloud.org/ns'
            },
            children: [{
                name: 'd:set',
                children: [{
                    name: 'd:prop',
                    children: []
                }]
            }]
}
```

PR #619, opened 2016-07-12
add StringUtility

georgehrke:


georgehrke:
test cases:
- [x] should always be lowercase
- [x] cover all code paths
- [x] test with emojis

PR #622, opened 2016-07-13
test autocompletion and settingsservice

georgehrke:


georgehrke:
please review @raghunayyar

PR #627, opened 2016-07-20
Fixing import of calendar files and show them directly after import is finished

tcitworld:


tcitworld:
Maybe `$rootScope.$broadcast('reloadCalendarList');` is necessary.

georgehrke:
What exactly does this fix?
Importing works just fine for me

tcitworld:
It stayed always at the "Analyzing..." step.

georgehrke:
okay, I will check.
Thx

js/app/controllers/importcontroller.js
```diff
@@ -122,7 +124,23 @@ app.controller('ImportController', ['$scope', '$filter', 'CalendarService', 'VEv
 				$scope.$apply();
 				$scope.closeIfNecessary();
 
-				//TODO - refetch calendar
+				CalendarService.getAll().then(function(calendars) {
```
georgehrke, line 127: This should really not be necessary. You are reloading all calendars once for each ics file the user selects.
Let me look into a better solution.

js/app/models/ImportFileWrapperModel.js
```diff
@@ -120,8 +120,8 @@ app.factory('ImportFileWrapper', function(Hook, SplitterService) {
 				context.progressToReach = context.splittedICal.vevents.length +
 					context.splittedICal.vjournals.length +
 					context.splittedICal.vtodos.length;
-				afterReadCallback();
 				iface.state = ImportFileWrapper.stateAnalyzed;
+				afterReadCallback();
```
georgehrke, line 124: This already fixes the issue for me.
Will drop the other changes from this commit.

tcitworld, line 124: Sure, but it would be nice for the imported calendar to show up directly without having to refresh the page.

georgehrke:
let's continue in https://github.com/owncloud/calendar/pull/629

Issue #628, opened 2016-07-20
Long filename during import results in off-screen button


fredfs456:
### Steps to reproduce
1. Have an ical file with a long (~60 characters) filename
2. Import through web UI by import > browse > select ,ics file
### Expected behaviour

-Filename is ellipsized or wrapped
-drop-down to select which calendar to import into (or new calendar) is visible
-button to start import is visible
### Actual behaviour

-Filename pushes drop-down and button off the right edge of dialog box
-Screenshot: http://i.imgur.com/ooLNekw.png
### Server configuration

Server configuration

Operating system:
Debian 8.5 Jessie
Web server:
Apache 2.4.10 (Debian)
Database:
MariaDB 10.0.25
PHP version:
5.6.22-0+deb8u1
ownCloud version: (see ownCloud admin page)
Nextcloud 9.0.53
Updated from an older ownCloud or fresh install:
Fresh install
### Client configuration

**Browser:**
Firefox 47
**Operating system:**
Windows 10

Issue #389, opened 2016-03-30
Calendar import doesn’t show calendar directly


jancborchardt:
Importing a calendar from an ICS file, I get the calendar title and the spinner in place of the color. That’s fine.

But then nothing really happens.

Out of curiosity I tried clicking on the calendar name, then clicked it again and the spinner was gone then and the events appeared. Really strange that it doesn’t do that automatically.

georgehrke:
1.0 or master?

jancborchardt:
master

georgehrke:
can't always reproduce, needs some more investigation, moving to 1.2

areteichi:
As I wrote in the other thread, this happens to me on ownCloud 9.0.1 and Calender 1.1 as well.

> I think I'm having the same issue.
> 
> I recently fresh installed ownCloud 9.0.1 and imported my calender data from ownCloud 8.2.3 (by exporting/importing them as ICS file). I've discovered that I can see all the events March 2016 onwards, but all the events prior to February 2016 seem not to be accessible.
> 
> The color icon on the left panel just keeps on spinning and the calender does not load when I go back, say, to February 2016.

PR #629, opened 2016-07-21
Fix import 2

georgehrke:
supersedes https://github.com/owncloud/calendar/pull/627

### DIFF
--- a/js/app/controllers/importcontroller.js
+++ b/js/app/controllers/importcontroller.js
@@ -80,6 +80,7 @@ app.controller('ImportController', ['$scope', '$filter', 'CalendarService', 'VEv
 						$scope.writableCalendars.push(calendar);
 					}
 					importCalendar(calendar);
+					fileWrapper.selectedCalendar = calendar.url;
 				});
 			} else {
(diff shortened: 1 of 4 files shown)


georgehrke:
@tcitworld Please check my latest commit.
This way we only reload the calendar in question :)

georgehrke:
fixes https://github.com/owncloud/calendar/issues/389

georgehrke:
@tcitworld this fixes https://github.com/owncloud/calendar/issues/628, right?

tcitworld:
Yup. See below. The filename span could do with a little margin-right, though (it sticks to the input next to it).
longfilename

georgehrke:
Okay, I added `...` to make clear a filename is cut off.

georgehrke:
Best solution would probably fixing this in CSS, but we can fix this later.

tcitworld:
You mean keeping the whole file name and using something like `text-overflow: ellipsis;` ?

georgehrke:
Yes, but let's get this in for now.
The CSS for the import dialog needs to be cleaned up anyway 🙈

Issue #599, opened 2016-06-29
Auto-refresh


ivnglkv:
I think it would be very useful feature for shared calendars.

georgehrke:
We should implement this in two steps:
- [ ] periodically check for updates of event in known calendars
- [ ] periodically check for new or updated calendars

PR #631, opened 2016-07-22
Webcal rebased

tcitworld:
Didn't want to pollute the webcal branch so here is my take on it https://github.com/owncloud/calendar/pull/443.
Remarks : 
- yes, there is an **awful lot** of common code in the Subscription and Calendar models. I need to find how to extend the Subscription model from Calendar properly. Sorry for not doing things properly from scratch. **Edit : Better now**
- subscription is not shown directly, calendar view is not reloaded, need to reload page
- when added, a subscription shows at first with the calendars
- Subscription events are editable
- There's a proxy and cache system for calendar files, and these files are stored inside the user's personal storage. Can it be otherwise (temporary files somehow) ? Hidden files ? Also, the cache time should be a setting (3600 hardcoded for now). Finally, files are saved with `basename(url)` as name, the subscription name should be better.
- No color choice for subscription (editable though)

TODOs
- [x] Refactor calendar/subscription model
- [x] Fix display of subscriptions : Possibly linked to #656
-  Make cache time a setting : Only admin can set this setting :/
- [x] Make subscription events read-only
- [x] Add color picker on creation
- [x] Use subscription name instead of `basename(url)` for the cache files names.
- [x] Tests !


appinfo/application.php
```diff
@@ -55,6 +55,16 @@ public function __construct($params=[]) {
...
 		});
+		$container->registerService('SubscriptionsProxyController', function(IAppContainer $c) {
+			$request = $c->query('Request');
+			$userSession = $c->getServer()->getUserSession();
+			$storage = $c->query('AppStorage');
+
+			return new Controller\SubscriptionsProxyController($c->getAppName(), $request, $userSession, $storage);
+		});
+		$container->registerService('AppStorage', function($c) {
+			return $c->query('ServerContainer')->getAppFolder();
```
georgehrke: If I recall correctly `->getAppFolder()` is really buggy.
@BernhardPosselt Didn't you mention smth like this?

BernhardPosselt: AFAIK its broken

georgehrke:
Let's please discuss how to proceed before you put more time into this. :)
I actually see quite a few issues in here.

appinfo/application.php
georgehrke: backlog, awesome 😞 
https://github.com/owncloud/core/issues/10767

controller/subscriptionsproxycontroller.php
```diff
@@ -0,0 +1,127 @@
...
+	/**
+	 * @NoAdminRequired
+	 * @NoCSRFRequired
+	 *
+	 * @param string $icsurl
+	 * @return DataDisplayResponse
+	 */
+	public function getIcsFile($icsurl) {
+		try {
+			$file = $this->storage->get('/' . basename($icsurl));
```
georgehrke: I'm generally not very happy about storing the webcal data on the server.
This should be really the last resort if we can't circumvent the CSP errors.

Additionally `getAppFolder` is on a per-app basis, not per-app and per-user.
The current way a user could access all webcal files stored on the server.

tcitworld: > I'm generally not very happy about storing the webcal data on the server.
> This should be really the last resort if we can't circumvent the CSP errors.

Well, I thought of the number of requests it might make for the distant server (especially if https://github.com/owncloud/calendar/issues/599 is active). Also, it's faster. But you can have just a proxy and serve directly the downloaded file.

> Additionally getAppFolder is on a per-app basis, not per-app and per-user.
> The current way a user could access all webcal files stored on the server.

Yup, will change.

js/app/models/subscriptionmodel.js
```diff
@@ -0,0 +1,245 @@
...
+ * You should have received a copy of the GNU Affero General Public
+ * License along with this library.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+app.factory('Subscription', ['Calendar', '$http', '$rootScope', 'Hook', 'VEventService', 'TimezoneService', function(Calendar, $http, $rootScope, Hook, VEventService, TimezoneService) {
+	'use strict';
+
+	function Subscription(url, props) {
+		// Calendar.apply(this, arguments);
```
georgehrke: If I'm not completely mistaken you can implement the hierarchy with `const iface = Calendar(url, props);`

js/app/service/calendarservice.js
```diff
@@ -260,6 +296,61 @@ app.service('CalendarService', ['DavClient', 'Calendar', function(DavClient, Cal
 		});
 	};
 
+	this.createSubscription = function(name, color, source) {
```
tcitworld, line 335: Makes more or less the same than `this.create`, just 2 little differences.

controller/subscriptionsproxycontroller.php
georgehrke: I wouldn't bother ;P
`SubscriptionsProxyController` is kind of problematic in my opinion.
It allows users to download any file (whatever size it has) and pollute the server's storage.

If we really want to cache it on the server, this should be done in the dav app and it should validate that it really only stores calendar-data.

@DeepDiver1975

js/app/filters/SubscriptionListFilter.js
```diff
@@ -29,10 +29,10 @@ app.filter('subscriptionListFilter', function(CalendarListItem) {
 		}
 
 		return calendarListItems.filter(function(item) {
-			if (!CalendarListItem.isCalendarListItem(item)) {
+			if (!CalendarListItem.isSubscriptionListItem(item)) {
```
georgehrke, line 32: You have to check for both calendarListItem and subscriptionListItem.
I have to admit that the naming is not the best here.

The Subscription Model is just about WebCal calendars.
Subscriptions in this filter are both webcal calendars and read-only calendars (like birthday-calendar and read-only shared-calendars)

I actually don't see a real need to distinguish between `CalendarListItem.IsCalendarListItem` and `CalendarListItem.isSubscriptionListItem`. What do you use this for?

controller/subscriptionsproxycontroller.php
tcitworld: >  It allows users to download any file (whatever size it has) and pollute the server's storage.

In order to fix that I can do a first HEAD request to get the file size and check it, then check it after it has been downloaded too. If the file is really too big (fake headers) then the handling will probably timeout anyhow.

> If we really want to cache it on the server, this should be done in the dav app

The CalDAV specification says subscriptions are only a list of remote urls, there's no handling of the files, therefore it would be only for this app.

js/app/models/subscriptionmodel.js
tcitworld: Thanks, thought it would be harder.

tcitworld:
Closing in favour of #694

PR #640, opened 2016-07-25
remove app.js from git

georgehrke:
remove app.js from git


georgehrke:
`grunt build` is already mentioned in https://github.com/owncloud/calendar/wiki/Developer-Info#building-javascript

PR #649, opened 2016-07-27
move some caldav calls into CalendarModel

georgehrke:


georgehrke:
let's put this on hold till public calendars are merged

hanzei:
Public calendars are merged. Can we merge this one now?

PR #652, opened 2016-07-29
move config to an unique route

tcitworld:


georgehrke:
- [x] please add a test for a not existing config key to make sure it returns the bad request

tcitworld:
I'll update https://github.com/owncloud/calendar/wiki/JSON-Settings-API later today.

Also, Scrutinizer seems to have issues.

Issue #636, opened 2016-07-24
Shared label overlaps calendar name, is not aligned, calendar 1.3


raimund-schluessler:
### Steps to reproduce
1. Share a calendar with a long name with another user
2. Have a look at the shared label
### Expected behaviour

Shared label should not overlap the calendar name, should be vertically aligned with the calendar name
### Actual behaviour

It overlaps and is not aligned, see attached screenshot

![sidebar](https://cloud.githubusercontent.com/assets/2496460/17084389/6a1d9718-51bc-11e6-8d2e-cbc25b4af75d.png)
### Server configuration

**ownCloud version:** 9.0.4

**Calendar version:** 1.3
### Client configuration

**Browser:** Firefox 47.0

**Operating system:** Win 10

PR #653, opened 2016-07-29
Fix vertical calendar alignment, get rid of inline css

raimund-schluessler:
As an addition to #636 I fixed the vertical alignment of the shared label and got rid of the inline css and the relative positioning. This also is the "somewhat more elegant solution" requested in https://github.com/owncloud/calendar/pull/571#discussion_r66419140 I think :wink: 

Also see the screenshots.

Before:
before

After:
after

### DIFF
--- a/css/calendarlist.css
+++ b/css/calendarlist.css
@@ -184,13 +184,29 @@
 	display: inline-block;
 }
 
+#app-navigation .app-navigation-list-item .utils span.action.withitems {
+	display: inline-block;
+	vertical-align: top;
+}
+
 #app-navigation .app-navigation-list-item .utils .action span {
 	display: inline-block;
-	line-height: 44px;
-	height: 16px;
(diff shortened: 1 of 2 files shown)


georgehrke:

jancborchardt:
@raimund-schluessler did you switch around the before & after screenshots? ;) Because the »Shared« text being close to the share icon is correct as they belong together.

raimund-schluessler:
@jancborchardt No, they are correct. This PR is about the vertical alignment, but fixing this lead to the padding aroud the spans being respected. I can fix the padding to get the label closer to the icon.

raimund-schluessler:
@jancborchardt I fixed the space between the icon and the label. See the screenshots and the Gif for clarification.

Before:
before

After:
after

Animated:
alignment

jancborchardt:
Perfect! :+1:

PR #658, opened 2016-08-01
Add labels to calendar dropdown menu

raimund-schluessler:
This PR adds labels to the dropdown menu of the calendar list:

Before:
before

After:
after

@jancborchardt @raghunayyar

### DIFF
--- a/css/calendarlist.css
+++ b/css/calendarlist.css
@@ -198,7 +198,6 @@
 	height: 36px;
 	line-height: 36px;
 	padding-right: 10px;
-	vertical-align: top;
 	font-weight: 400;
 	float: left;
 }
--- a/css/calendarlist.css
+++ b/css/calendarlist.css
@@ -180,6 +180,35 @@
 	width: 75%;
 }
 
+#app-navigation .app-navigation-entry-menu li {
+	width: auto !important;
+	float: inherit;
+	height: 36px;
+}
+
+#app-navigation .app-navigation-entry-menu li button {
+	float: inherit !important;
+	margin: 0;
+	padding: 0px;
+	width: 100% !important;
+}
+
+#app-navigation .app-navigation-entry-menu li span {
+	display: inline-block;
+	height: 36px;
+	line-height: 36px;
+	padding-right: 10px;
+	vertical-align: top;
+	font-weight: 400;
+	float: left;
+}
+
+#app-navigation .app-navigation-entry-menu li span.svg {
+	padding: 5px;
+	width: 36px;
+	box-sizing: border-box;
+}
+
 #app-navigation .app-navigation-list-item .utils {
 	display: inline-block;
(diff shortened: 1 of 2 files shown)


css/calendarlist.css
```diff
@@ -180,6 +180,35 @@
...
+#app-navigation .app-navigation-entry-menu li {
+	width: auto !important;
+	float: inherit;
+	height: 36px;
+}
+
+#app-navigation .app-navigation-entry-menu li button {
+	float: inherit !important;
+	margin: 0;
+	padding: 0px;
```
tcitworld, line 192: `px` is not needed here.

css/calendarlist.css
```diff
@@ -180,6 +180,35 @@
...
+	width: 100% !important;
+}
+
+#app-navigation .app-navigation-entry-menu li span {
+	display: inline-block;
+	height: 36px;
+	line-height: 36px;
+	padding-right: 10px;
+	vertical-align: top;
```
tcitworld: `vertical-align` doesn't work with `float: left`

georgehrke:
Maybe `Download` instead of `Export`? cc @jancborchardt

css/calendarlist.css
tcitworld: (i.e : is not needed)

MariusBluem:
Hmmm... We are also calling it `import` instead of `upload` 😕 @georgehrke 
We should call it `Link` instead of `CalDAV-Link` ... Sounds too technical 😁

tcitworld:
Import makes a sense since you can import events from an .ics file into an existing calendar.

georgehrke:
- [ ] test with different locals

raimund-schluessler:
@tcitworld Thanks for the feedback on the CSS, I will change it accordingly.

>  test with different locals

Will do, but it should work as expected because the width is adjusted automatically.

Regarding the text of the labels I just used the text which was used for the titles. We should keep that consistent.

So is this the current agreement? @jancborchardt 
- Edit
- Link
- Download
- Delete

georgehrke:
I delegate my vote regarding the labels to @jancborchardt

raghunayyar:
@georgehrke we should get it in, we will change the name if @jancborchardt suggests so. Have you tested on different locals? :)

georgehrke:
no, have not tested different locals yet, feel free to do so

jancborchardt:
Changed the wording to »Download« and »Link« at https://github.com/owncloud/calendar/pull/695, please review. :)

@raimund-schluessler great work! :) Do you want to fix that in the Mail app as well maybe?

jancborchardt:
Ideally we should port the styles over to the apps.css to have them in the main code. cc @mar1u5 @raimund-schluessler if you want to take that up. :)

PR #661, opened 2016-08-02
Add css lint to travis.

tcitworld:
Yes, I like to lint. Your opinion ?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -66,6 +66,7 @@ script:
     - phpunit --coverage-clover=coverage.clover tests/unit
     - cd js/
     - grunt karma
+    - ./node_modules/stylelint/dist/cli.js ../css/*.css
     - cd ..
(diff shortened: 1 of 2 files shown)


raghunayyar:
:+1:

tcitworld:
Just need to test that the build fails correctly.

tcitworld:
It's fine.

georgehrke:
Would it be possible to add a grunt command for `./node_modules/stylelint/dist/cli.js ../css/*.css` ?

tcitworld:
There already is one : `grunt postcss` (I anticipated eventual extra uses of postcss). However right now it doesn't return a proper failure, it just shows warnings. 
Maybe it can be configured instead of calling stylelint this way. But we can also get rid of postcss and use `grunt-stylelint` directly.

PR #663, opened 2016-08-02
Fix on hover flickering share delete icon

raimund-schluessler:
This PR fixes the wrongly positioned share delete icon which jumped to the correct position on hover.

Before:
before

After (or before while hovering):
after

### DIFF
--- a/css/app/calendarlist.css
+++ b/css/app/calendarlist.css
@@ -238,8 +238,8 @@
 	padding-right: 5px;
 }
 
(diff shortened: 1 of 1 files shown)


raimund-schluessler:
Maybe @raghunayyar ?

georgehrke:
awesome, this flickering always bugged me like hell!
Will test asap

georgehrke:
rebased ...

tcitworld:
:+1:

Issue #664, opened 2016-08-02
Timezone "None" doesn't work


wrexroad:
### Steps to reproduce
1. Create a new event by clicking a date on the calendar.
2. Click the "More" option.
3. Select global timezone "None".
### Expected behaviour

Who knows? "None" is pretty ambiguous for a timezone.
### Actual behaviour

TimezoneService tries to get the listing for the "FLOATING" timezone:

```
GET [OWNCLOUDSERVER]/apps/calendar/v1/timezones/FLOATING.ics 404 (Not Found)
```

The 404 response is the processed as a timezone and subsequent selections of "None" result in a broken promise. Tears and heartbreak are sure to follow:

```
Uncaught (in promise)
  Object {data: "<!DOCTYPE html>↵<!--[if lte IE 8]><html class="ng-…ll your data            </p>
  </footer>↵  </body>↵</html>↵", status: 404, config: Object, statusText: "Not Found"}

```
### Server configuration

**Operating system**:
Ubuntu 14.04
**Web server:**
Apache
**Database:**
MySQL
**PHP version:**
5.5.9
**ownCloud version:** (see ownCloud admin page)
Nextcloud 9.0.53 (stable)
**Calendar version:** (see ownCloud apps page)
1.3.1 (Also tested on GitHub master branch commit 8addb4).
**Updated from an older ownCloud or fresh install:**
Freshly built from GitHub repo.
**Signing status (ownCloud 9.0 and above):**

```
No errors have been found.
```
### Client configuration

**Browser:**
Chrome
**Operating system:**
Fedora 24
#### Browser log

See Above

georgehrke:
We should probably rename it to floating though

PR #665, opened 2016-08-02
Remove tz none

wrexroad:
Fixes issue #664 

Looking at Mozilla's ical.js, it looks like they intended "floating" to indicate the local timezone, not no timezone. Since the local timezone is already automatically detected and selected it makes sense to just remove the "None" options.


georgehrke:
Floating timezones has other benefits. It "travels" with you.
Also see https://tools.ietf.org/html/rfc5545#section-3.3.5

Removing it completely is no solution 👎

wrexroad:
Ah ha! I didn't fully appreciate what floating was supposed to do!

PR #667, opened 2016-08-02
Fix global tz

wrexroad:
Fixes issue #666.
After some investigation when the `EditorController` tries to save the event, `SimpleEvent` tries to process the dates in `SimpleEvent#specificReader.date()`. As part of this process, it checks which timezones components have already been created for the event and stores them in `availableTimezones`. Then, if the timezone of the date that is being added is not in that list of existing timezones, it is fetched from ICAL.TimezoneService and the component is added.

This is problematic because ICAL.Timezone.utcTimezone does not have a `component` property so running `vevent.parent.addSubcomponent(startTz.component);` or `vevent.parent.addSubcomponent(endTz.component);` throws an exception.

The solution would seem to be to avoid adding global timezone components by initializing with `var availableTimezones = ['UTC', 'GMT', 'Z']`. While this does allow the event to be created, the ical file has the start and stop times listed as: DTSTART:TZID=UTC;[YYYYMMDD]T[HHMMSS]

With the TZID parameter still being specified we need a VTIMEZONE property which is not being provided. We could create the component manually, but a better option would be to remove the TZID altogether and have the date-time string itself indicate it is UTC with a 'Z' appended on the end:  DTSTART:[YYYYMMDD]T[HHMMSS]Z.

To get the correct string output, the `useUtc` flag needs to be set when creating dates with `ICAL.Time.fromJSDate`. This does indeed gives the correct UTC formatted string, but it gives the wrong time. It turns out this is because of how `moment.js` outputs dates with `toDate`. Interestingly, if you try `moment.utc().toDate()` you will get a local date, NOT a UTC date.

Instead of using `ICAL.Time.fromJSDate` we can just use `ICAL.Time.fromData` and provide an object which describes the date but is timezone agnostic. Then we pass in the correct timezone as a second parameter.

The timezone is set with:

```
var startTz = newSimpleData.dtstart.parameters.zone === "floating" ? ICAL.Timezone.localTimezone : ICAL.TimezoneService.get(newSimpleData.dtstart.parameters.zone);
```

This will catch any floating timezones and set them to localtime. It will also correctly return ICAL.Timezone.utcTimezone when a global time is set.

Finally, the TZID parameter is only set if the ICAL provided timezone is not 'UTC'.


georgehrke:
works 👍 
thx for your contribution :)

georgehrke:
I'm afraid we have to revert this merge.
This introduces a severe bug:
- when selecting a time range the generated event has an offset in its time

wrexroad:
Can you elaborate a little more on the error you are getting? Here are the steps I used to test:
1. Click empty date and create a new event (8/15/2016 in this test).
2. Click "More" to get the sidebar to pop out.
3. Uncheck "All day event".
4. Click Timezone button.
5. Select UTC.
6. Click "Create".
7. Open the event sidebar again and export the .ics file.

The `VEVETN` component of the .ics file is:

```
BEGIN:VEVENT 
CREATED:20160812T144142
DTSTAMP:20160812T144142
LAST-MODIFIED:20160812T144142
UID:i0z5t1w0kftulhvoksex6flxr
SUMMARY:utc
STATUS:CONFIRMED
DTSTART:20160815T000000Z
DTEND:20160815T010000Z
END:VEVENT
```

Similar results for converting an existing event with a local timezone to UTC.

What am I missing?

georgehrke:
When I select the dates 15th thru 16th, it creates an event from 14th to 15th.
What timezone are you in?

wrexroad:
I'm in California and we are currently UTC-0700. When I create an event using the default time (12AM - 1AM) on the 15th and set UTC timezone, it is displayed on the calendar on the 14th because the calendar is in my local time (12AM UTC on the 15th is 5PM on the 14th in California). I just switched my computer to berlin time (UTC+0200) and the 12AM UTC event on the 15th shows up as 2AM on the 15th. So it looks like everything is working correctly there.

wrexroad:
Just out of curiosity when you create an event with timezone `Danmarkshavn (America)` do you see the same result? They are currently in Daylight Savings Time which pushes their offset to `UTC+0000`. So it should create a pseudo-UTC event which does not use the code in this pull request. For me both the `UTC` timezone in this pull request and the `Danmarkshavn (America)` timezone create identical time ranges.

georgehrke:
Sorry, I was not very clear in my last comment. The problem happens not when saving the issue, but already when selecting a time range in the calendar.

I selected the days 15th thru 16th, but the editor contained the days 14th thru 15th.

wrexroad:
Oh, interesting! I think I have found our disconnect. I have been doing all of my testing using the month view, which does not exhibit the behavior you are describing. However, when I use the week or day view the time is initialized without the TZ offset and the editor times/dates are messed up. I will see what I can do to fix this.

georgehrke:
I can reproduce it in month view as well, but i think the reason is simple: You are in UTC-xx and I'm and in UTC+xx.

wrexroad:
That sounds like a reasonable possibility, either way I think I understand the root cause. Hopefully I can get an updated fix later this week.

Issue #395, opened 2016-03-30
scrolling in sidebar broken


georgehrke:
not possible to see everything

georgehrke:
@raghunayyar Do you have time to look into this? :)
If the window's height is not big enough the buttons overlay the content which is not scrollable

georgehrke:
@raghunayyar Mind taking a look?

onny:
There seems to be some styling issues in the sidebar ...
![2016-05-21-125934_swaygrab](https://cloud.githubusercontent.com/assets/757752/15448167/7022fb6a-1f5a-11e6-9dcc-c5532be608d4.png)

georgehrke:
@onny That's the app-navigation. With sidebar I'm referring to the advanced events editor on the right. 
Please report a dedicated bug about the app-navigation issue :)

onny:
So you mean this styling issue, right?
![right_menue](https://cloud.githubusercontent.com/assets/757752/17189039/0a4eab20-5440-11e6-9928-0410a5f54d48.png)

georgehrke:
exactly

MariusBluem:
Thats what we've spoken about @jospoortvliet @jancborchardt :wink:

oparoz:
It makes it impossible to change the reminders

georgehrke:
@raghunayyar I'd appreciate if you could take a look at this very soon

raghunayyar:
Yes, on it.

raghunayyar:
Fixed, please check guys. :)

steckerhalter:
@raghunayyar Seems like the overflow is hidden and there is no scrollbar in Calendar v1.3.3:

![](https://cloud.githubusercontent.com/assets/3024024/18578494/8d43d136-7bf0-11e6-9d09-415774ea493f.png)

So calendars below are not visible at all.

georgehrke:
The sidebar is referring to the advanced events editor.
Please report a dedicated issue

PR #670, opened 2016-08-03
fix display of details event sidebar on low resolutions

tcitworld:
Buttons were shown in the middle of the view on low resolutions
event-details-lowres

However, on high res, there are no longer displayed totally on the bottom, so please tell me if that's an issue.
event-details-highres

### DIFF
--- a/css/app/eventdialog.css
+++ b/css/app/eventdialog.css
@@ -117,13 +117,8 @@
 }
 
 .advanced--container {
-	position: absolute;
-	top: 0;
-	left: 0;
-	right: 0;
-	bottom: 0;
 	padding: 20px;
-	margin-bottom: 120px;
+	margin-bottom: 40px;
 }
 
 .advanced--fieldset {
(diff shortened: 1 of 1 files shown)


georgehrke:

raghunayyar:
On low resolution I see the screen as this. The fixed bottom overlaps / too tight with the content above it. Can you check?

screen shot 2016-08-04 at 3 38 34 pm

georgehrke:
@raghunayyar That's actually another issue

georgehrke:
One that's assigned to you ;P
https://github.com/owncloud/calendar/issues/395

tcitworld:
Well, it's related, I was trying to fix this behavior too.

jancborchardt:
@georgehrke @raghunayyar why is there such a giant export button btw? And the close button should just be an X in the top right?

georgehrke:
created an issue about the close button: https://github.com/owncloud/calendar/issues/681

georgehrke:
> why is there such a giant export button btw?

When the cancel button is gone we can just put it next to the delete button

raghunayyar:
This issue is already resolved in my PR. Thanks. :-)

PR #671, opened 2016-08-03
Predefine the "None" timezone

wrexroad:
Fixes issue #664 
Rather than removing "None", TimezoneService defines "FLOATING" as ICAL.Timezone.localTimezone similarly to how it defines the UTC time zones. This prevents the attempt to download the FLOATING .ics file which does not exist.

### DIFF
--- a/js/app/service/timezoneservice.js
+++ b/js/app/service/timezoneservice.js
@@ -31,6 +31,7 @@ app.service('TimezoneService', ['$rootScope', '$http', 'Timezone', 'TimezoneList
 		this._timezones.UTC = new Timezone(ICAL.TimezoneService.get('UTC'));
 		this._timezones.GMT = this._timezones.UTC;
 		this._timezones.Z = this._timezones.UTC;
+		this._timezones.FLOATING = new Timezone(ICAL.Timezone.localTimezone);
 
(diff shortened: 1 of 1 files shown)


georgehrke:
make sense, although this does not fix the wrong behavior in the simpleEventModel

georgehrke:
thx for the contribution! :)

wrexroad:
@georgehrke, what is the wrong behavior in simpleEventModel? As far as I can tell things look like they are working fine there with the floating timezone.

georgehrke:
I haven't checked, but does it remove existing TZIDs?

wrexroad:
I just double checked by creating an event in my timezone, exporting it, then updating it to timezone None and exporting it. All of the TZID parameters on DTSTART and DTEND were removed. The only timezone info left is the VTIMEZONE component from my timezone which seems to be created regardless of the event timezone. The superfluous VTIMEZONE wont hurt anything and is probably part of a different issue anyway..

georgehrke:
okay, thx for checking :)

georgehrke:
I created an issue about the superfluous VTimezones https://github.com/owncloud/calendar/issues/674

PR #677, opened 2016-08-04
add popoverPositioningUtility

### LINKED ISSUE #587: move CalController::_calculatePopoverPosition to a PopoverPositioningUtility

georgehrke:
fixes https://github.com/owncloud/calendar/issues/587

### DIFF
--- a/js/app/controllers/calcontroller.js
+++ b/js/app/controllers/calcontroller.js
@@ -26,8 +26,8 @@
 * Description: The fullcalendar controller.
 */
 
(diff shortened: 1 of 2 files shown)


georgehrke:
Ideally we would implement this with CSS classes only.

This has some issues just like master:
- Popover doesn't reposition on window resize

georgehrke:
please review @raghunayyar

raghunayyar:
Agreed on the CSS part, but lets just get this is in as of now.

PR #680, opened 2016-08-05
get rid of angular ui calendar

### LINKED ISSUE #201: try to replace all the rootscope broadcasts with $scope.$watchCollection()

### LINKED ISSUE #514: get rid of angular-ui-calendar
- we can write our own directive
- angular-ui-calendar has old dependencies

georgehrke:
fixes https://github.com/owncloud/calendar/issues/514
also fixes https://github.com/owncloud/calendar/issues/201


georgehrke:
angularUICalendar was apparently only a proof of concept and was never meant to be maintained for a longer period.
it comes with crazy old dependencies and has lots of features we don't use

this introduces our own directive for fullcalendar

georgehrke:
please review @raghunayyar

raghunayyar:
Failing tests, can you see?

raghunayyar:
After I create an event, and click on any other day, it creates an event without opening the events editor. This goes on forever. Can you check @georgehrke ?

georgehrke:
can't reproduce.
Did you remember to build the JS to test?

raghunayyar:
This is weird, it works. :)

wrexroad:
Im having a similar problem to @raghunayyar. I can create an event, but any subsquent clicks just "creates" more events without opening the editor. These events aren't actually created though as they disappear on page reload.
Each click also throws the following error:

```
app.js?v=3ffa53a…:4404 Uncaught TypeError: Cannot read property 'dismiss' of null
```

Looks like the error is coming from `eventseditordialogservice.js`, the block at like 130.

georgehrke:
@wrexroad Should have been fixed with https://github.com/owncloud/calendar/pull/685 :)

wrexroad:
Ah! It looks like you merged that in just after I pulled down a current version of the repo.

Issue #674, opened 2016-08-04
Cleanup VTimezones when editing events


georgehrke:
When a user changes an event's timezone the old VTimezone information is not removed

PR #683, opened 2016-08-05
Added cleanup vtimezone cleanup routine.

wrexroad:
Fixes issue #674  
After `EditorController` validates the event and patches in the changes, it checks all non-timezone subcomponents of the `vevnet` variable for `TZID` parameters. Then it checks each `VTIMEZONE` subcomponent to see if its TZID was referenced by any of the non-timezone subcomponents. If it has not been references at all the timezone is removed from `vevent`.

### DIFF
--- a/js/app/controllers/editorcontroller.js
+++ b/js/app/controllers/editorcontroller.js
@@ -103,6 +103,7 @@ app.controller('EditorController', ['$scope', 'TimezoneService', 'AutoCompletion
 
 			$scope.prepareClose();
 			$scope.properties.patch();
+			$scope.cleanupTimezones();
 			$uibModalInstance.close({
 				action: 'save',
 				calendar: $scope.calendar,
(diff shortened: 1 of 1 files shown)


georgehrke:
I feel like this should be moved into the simpleEventModel. What do you think?

wrexroad:
`SimpleEventModel` was my first thought, but the main problem is that it only represents the `VEVENT` subcomponent, and `tzid`'s can potentially be defined elsewhere. Furthermore, `VTIMEZONE` is a sibling of  `VEVENT`, not a child. So if `SimpleEventModel` represents `VEVENT` it shouldn't be its job to clean up the sibling components.

It looks like `VEventModel` is what actually represents the `VCALENDAR` supercomponent. So maybe it would make sense to move it into there, but I didn't see a place that it would fit nicely.

Issue #452, opened 2016-04-12
more sharing dialog css fixes needed


georgehrke:
https://github.com/owncloud/calendar/pull/412#issuecomment-204920448

Update:-
- [x] when typing a username the input field "jumps down" a few pixels until the results are loaded
- [x] There is some annoying whitespace in the bottom of the dialog.

PR #689, opened 2016-08-08
Whitespace fix, no jumping of dialog.

raghunayyar:
Fixes both of these issues with our sharing dialog. https://github.com/owncloud/calendar/issues/452


georgehrke:
👍

PR #693, opened 2016-08-10
speed up event rendering for individual calendars

georgehrke:


georgehrke:
please test @raghunayyar

Issue #364, opened 2016-03-23
Cache webcal calendars in browser's local storage


georgehrke:
in addition to: https://github.com/owncloud/calendar/issues/132

cache (not password protected) webcal subscriptions using HTML5 Browser Storage.

needs discussion about security:
- users with the same browser profile might be able to read the locale storage. Potentially dangerous when webcal subscription contains private data. (e.g. Facebook provides a webcal (url contains hash) with the events you attend or you saved)
  - can we add something like "on logout" scripts to clear locale storage when a user logs out?
  - Is there some easy way to encrypt the data?

georgehrke:
cc @LukasReschke

DeepDiver1975:
this way any webcal is only available in the web app - I'd expect them in any connected caldav client as well. can subscription help us with this respect?

georgehrke:
As far as I understood CalDAV subscriptions its only meant to store the actual link and not the objects itself. Is this wrong?

DeepDiver1975:
> As far as I understood CalDAV subscriptions its only meant to store the actual link and not the objects itself. Is this wrong?

I was hoping you already know more about this :sob: :wink:

georgehrke:
it's already pretty hard to find some documentation about it.
Can you link me to the standard?

georgehrke:
According to [this sabre file](https://github.com/fruux/sabre-dav/blob/master/lib/CalDAV/Subscriptions/Plugin.php#L13) it's only a list.

Searching the web for `CalDAV managed subscriptions` or `"{http://apple.com/ns/ical/}refresh rate"` didn't turn up anything useful :confused: 

I begin to question this standard's existence :sob:

georgehrke:
https://github.com/fruux/sabre-dav/blob/master/tests/Sabre/CalDAV/Subscriptions/CreateSubscriptionTest.php#L19L39

Although this still doesn't explain whether or not the server is supposed to cache the objects or if it's really just a list of subscriptions

georgehrke:
Maybe @evert can enlighten us :laughing:

evert:
It's just a list of subscriptions. Although a server-side cache might be cool =)

I don't think a spec exists. I believe I just reverse-engineered it.

georgehrke:
> I don't think a spec exists.

Haha, that explains a lot :see_no_evil: 

Thx a lot :)

georgehrke:
[...]

PR #694, opened 2016-08-10
implement WebCal support

georgehrke:


phsc84:
I'm just curious: What's the difference to #631?

georgehrke:
It uses a different approach.
#631 caches the ics files on the server. I'm not really happy with that approach.
This branch handles all of this in the user's browser.
It also reuses lots of code from #631 though.

georgehrke:
- [x] proper error handling
- [ ] tests 

~~[ ] [web-based protocol handlers](https://developer.mozilla.org/en-US/docs/Web/API/Navigator/registerProtocolHandler)~~

js/app/models/webcalModel.js
```diff
@@ -0,0 +1,81 @@
...
+ *
+ */
+
+app.factory('WebCal', function($http, Calendar, VEvent, TimezoneService, WebCalService) {
+	'use strict';
+
+	function WebCal(url, props) {
+		const context = {};
+
+		if (props.href.startsWith('webcal://')) {
```
tcitworld: Is this needed anymore with WebCalUtility ?

georgehrke: No, it is yet to be replaced

tcitworld:
As of now, remote calendars are not cached, correct ?

EDIT : Inside the browser, I mean.

georgehrke:
Exactly, it's not cached in the browser. That's what this ticket is about: https://github.com/owncloud/calendar/issues/364

(We store it in the WebCalService though. When you change month it doesn't query the webcal file again)

georgehrke:
- [x] fix download link to webcal calendars

http/StreamResponse.php
```diff
@@ -0,0 +1,51 @@
...
+	 */
+	public function __construct ($stream) {
+		$this->stream = $stream;
+	}
+
+
+	/**
+	 * @param IOutput $output a small wrapper that handles output
+	 */
+	public function callback (IOutput $output) {
```
tcitworld: What about if someone provides a huge file ?

georgehrke: That's why I decided to use streams. 😉

georgehrke:
- [x] `webcals` is not a thing. When URL is stored as `webcal` try `https` first and fall back to `http`

georgehrke:
- [x] don't show export button for events when accessibleViaCalDAV is false

tcitworld:
I got `Class 'OCA\Calendar\Http\StreamResponse' not found` for some reason.

tcitworld:
Add an optional field name for the subscription (because `http://www.mydom...`) is not a fancy name.

georgehrke:
> Add an optional field name for the subscription (because http://www.mydom...) is not a fancy name.

Hm, you can still rename the subscription after creating it. The point is to keep it as simple as possible and most WebCal files come with a predefined name.

tcitworld:
I'm fine with that.

js/app/models/webcalModel.js
```diff
@@ -0,0 +1,92 @@
...
+
+		iface.eventsAccessibleViaCalDAV = function() {
+			return false;
+		};
+
+		return iface;
+	}
+
+	WebCal.isWebCal = function(obj) {
+		return (typeof obj === 'object' && obj !== null && obj._isAWebCalObject === false);
```
tcitworld: Shouldn't this be `obj._isAWebCalObject === true` ?

georgehrke: Yes, absolutely.
I've no idea why it's false 🤔

tcitworld: Will change.

tcitworld:
@georgehrke Please have a look. I put a 2 hours cache timeout, but I guess that could be a setting.

tcitworld:
Do you have some public webcal (not just public .ics) resources in order to test the protocol handler ?

tcitworld:
- [ ] Add color picker

georgehrke:
Just like the Name I would not offer to enter a color at creation. Most webcal calendars have a color included.

georgehrke:
Cc @jancborchardt ^

tcitworld:
Facebook events webcal urls don't seem to work.

georgehrke:
I will check, but not before Monday afternoon.

tcitworld:
2 little issues : 
- when adding an url, the new events don't show up directly. You have to click for instance somewhere on the calendar view to have them show up.
- when adding an url and after the events have been shown, if you click on an event to see it's details, all the events in the Calendar view get the same title.

georgehrke:
> - when adding an url, the new events don't show up directly. You have to click for instance somewhere on the calendar view to have them show up.

Fixed

> - when adding an url and after the events have been shown, if you click on an event to see it's details, all the events in the Calendar view get the same title.

Not yet fixed, but I'm certainly sure I know why this happened. We don't assign a uri to webcal events yet.

georgehrke:
subscribing to the fb events webcal feed works just fine for me :/

georgehrke:
- [x] lock webcal url input after clicking button and show some spinner

tcitworld:
> subscribing to the fb events webcal feed works just fine for me :/

My bad, it seems to be an issue with my Facebook generated ics (Friends birthdays)...

```
Error: invalid date-time value: "2016--0-9-T6T::::"
```

EDIT : Facebook events feed works correctly.

Other bugs correctly fixed.
- [ ] Clear input field after adding the calendar

tcitworld:
- [x] Give original calendar link through calendar item menu.

tcitworld:
- [x] Handle cases where ressource is password protected, resulting in HTTP 401 and GuzzleHttp sending an exception.

tcitworld:
I passed the eventual http errors though the stream response own error code, but it may bring problems, so maybe through the body is enough.

georgehrke:
There is one important thing missing: When a server returns a 301, update the subscription url stored in the server. But that's something I would like to do in a follow up PR.

tcitworld:
Yeah I thought of that but forgot to report. Any service you know of to test with ?

georgehrke:
Just link an webcal feed via http://bit.ly/ 😂

georgehrke:
- [x] test exception in proxy controller

jancborchardt:
Yep, at creation it should simply be imported. Any changes to name or color can be made afterwards. But import should be as quick as possible.

georgehrke:
Let's do the web-based protocol handlers thing in a follow up pr. This pr is already huge

georgehrke:
protocol handlers: https://github.com/owncloud/calendar/issues/742

georgehrke:
the 301 thing: https://github.com/owncloud/calendar/issues/743

georgehrke:
There is only one thing left:
- [x] fix js unit tests

georgehrke:
@tcitworld I think this can go in. What do you think?

tcitworld:
I didn't test your last changes, but it should be good.

georgehrke:
okay :) 
Let's get this in!

tcitworld:
I'm getting php notice messages like `rewind(): stream does not support seeking` so [this line](https://github.com/owncloud/calendar/blob/master/http/streamresponse.php#L47) might now be needed (see http://stackoverflow.com/a/6518288).

georgehrke:
Hm, okay. I guess we can just remove the rewind after all

PR #695, opened 2016-08-11
fix wording in menu and remove unnecessary titles

jancborchardt:
Follow-up to https://github.com/owncloud/calendar/pull/658

Fixing the naming, and the title attribute is not necessary since it’s directly there as text. :)

Please review @raimund-schluessler @mar1u5 @georgehrke @raghunayyar


georgehrke:
works great, thx :)
👍

PR #697, opened 2016-08-13
make it possible to add prefix and suffix for StringUtility.uid

georgehrke:
please review @raghunayyar


raghunayyar:
Nice, merge when the build passes.

georgehrke:
Scrutinizer hangs, but it's not important -> merge

PR #701, opened 2016-08-16
Show random color button

### LINKED ISSUE #531: show random color button for ownCloud 9.1+

georgehrke:
reintroduce random color button
fixes https://github.com/owncloud/calendar/issues/531

Needs to be tested on:
- [x] stable9
- [x] stable9.1
- [x] master

### DIFF
--- a/tests/js/unit/utility/colorUtilitySpec.js
+++ b/tests/js/unit/utility/colorUtilitySpec.js
@@ -96,10 +96,10 @@ describe('ColorUtility shared tests', function () {
 
 	it('should call forward the call to hslToRgb', function () {
 		expect(ColorUtility._hslToRgb(6, 70, 68));
-		expect(hslToRgb).toHaveBeenCalledWith(6, 70, 68);
+		expect(hslToRgb).toHaveBeenCalledWith(6, 0.7, 0.68);
 
 		expect(ColorUtility._hslToRgb([6, 77, 68]));
-		expect(hslToRgb).toHaveBeenCalledWith(6, 77, 68);
+		expect(hslToRgb).toHaveBeenCalledWith(6, 0.77, 0.68);
 	});
 });
 
--- a/css/app/calendarlist.css
+++ b/css/app/calendarlist.css
@@ -273,7 +273,6 @@
 
 .colorpicker .colorpicker-list {
 	display: flex;
-	flex-wrap: wrap;
 	width: 100%;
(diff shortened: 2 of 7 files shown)


tests/js/unit/utility/colorUtilitySpec.js
```diff
@@ -96,10 +96,10 @@ describe('ColorUtility shared tests', function () {
 
 	it('should call forward the call to hslToRgb', function () {
 		expect(ColorUtility._hslToRgb(6, 70, 68));
-		expect(hslToRgb).toHaveBeenCalledWith(6, 70, 68);
+		expect(hslToRgb).toHaveBeenCalledWith(6, .7, .68);
```
tcitworld: JSHint shows warnings and the grunt tasks fails, so you should add a leading 0 I guess.

> A leading decimal point can be confused with a dot

tcitworld:
When creating a new calendar, the random color is not shown on the same line (french screenshot, but the same with en)
random_color

georgehrke:
@raghunayyar Please take a look at this new line issue mentioned by @tcitworld

raghunayyar:
Sure, will have a look at this asap.

raimund-schluessler:
@raghunayyar You can have a look at the Tasks app. There it is aligned correctly.

georgehrke:
@raimund-schluessler He already pushed a commit that aims to fix it :)
Haven't tested it yet.

georgehrke:
agenda - nextcloud chromium today at 11 44 33 am
Aligning issue seems to be fixed :)

I'll now test this on the other branches

georgehrke:
Tested on all supported server versions 👍

skjnldsv:
Well done!! :+1:

EtlamGit:
No, not well done!
The line with `flex-wrap` was intentionally placed there. If missing it destroys the layout when custom colors are defined. See the PullRequest #602 as reference.

The problem shown here looks like "too many" colors are defined to fit into one row. Therefore it wraps to the next row (so an intended behavior). A "good" fix would do one of these things:
- reduce the number of colors
- make the colors boxes a little bit smaller
- reduce the left and right margin
- widen the complete navigation pane
- remove the random button

PR #734, opened 2016-09-06
update dependencies

georgehrke:


georgehrke:
I was not able to find any regressions :)

tcitworld:
Maybe this would be a good opportunity to handle all dependencies though mpm (since an user isn't supposed to install through git anyway).

georgehrke:
> Maybe this would be a good opportunity to handle all dependencies though mpm

Why would you ditch bower? 

Or do you just mean not to check in the code of dependencies? 
I think it has some benefits, because if the developer unpublishes it, we still have the code in a safe place, but it's definitely something to discuss :)

tcitworld:
I don't mean to ditch bower, just move the libraries currently checked in to npm. npm certainly has issues, for instance packages disappearing, but I find it so sad to have such commits just to bring new versions for the dependencies.
Since the release packages will contain the dependencies anyway, I would say we're relatively safe.

georgehrke:
I hardcoded the versions in the bower.json now :)

tcitworld:
Great. In fact, bower is quite enough for what we're doing here (the Makefile handles the rest).

georgehrke:
I was kind of expecting travis to fail 🙈

georgehrke:
I will look into the failing travis tests later today, will probably just fix https://github.com/owncloud/calendar/issues/735 to get this resolved

tcitworld:
Fixed travis, but this can certainly be improved.

georgehrke:
thx :) 

Will work on #735 soon

Issue #730, opened 2016-09-02
Month name is shown as "undefined"


vedranmiletic:
### Steps to reproduce
1. Open the Calendar app.
2. Use the Month view.
### Expected behaviour

Month name should be displayed correctly, e.g. "September 2016".
### Actual behaviour

Month name is shown as "undefined 2016".
### Server configuration

**Operating system:**
Fedora 25

**Web server:**
Apache 2.4.23

**Database:**
PostgreSQL 9.5.4

**PHP version:**
7.0.10

**Server version:** (see your admin page)
Nextcloud 10.0 (stable)

**Calendar version:** (see the apps page)
Calendar 1.3.3

**Updated from an older installed version or fresh install:**
Updated from Nextcloud 9.0.53

**Signing status (ownCloud/Nextcloud 9.0 and above):**
No errors have been found.

**List of activated apps:**
Enabled:
- activity: 2.3.2
- admin_audit: 1.0.0
- bookmarks: 0.9.0
- calendar: 1.3.3
- comments: 1.0.0
- contacts: 1.3.1.0
- dav: 1.0.0
- documents: 0.13.1
- federatedfilesharing: 1.0.1
- federation: 1.0.1
- files: 1.5.2
- files_pdfviewer: 0.8.1
- files_sharing: 1.0.0
- files_texteditor: 2.1
- files_trashbin: 1.0.0
- files_versions: 1.3.0
- files_videoplayer: 0.9.8
- firstrunwizard: 1.1
- gallery: 15.0.0
- notifications: 0.3.0
- password_policy: 1.0.0
- provisioning_api: 1.0.0
- serverinfo: 1.1.1
- survey_client: 0.1.5
- systemtags: 1.0.2
- templateeditor: 0.1
- theming: 1.0.1
- updatenotification: 1.0.1
- workflowengine: 1.0.1
  Disabled:
- encryption
- external
- files_accesscontrol
- files_automatedtagging
- files_external
- files_retention
- user_external
- user_ldap
- user_saml

**The content of config/config.php:**
{
    "system": {
        "instanceid": "ocrme1da2m3y",
        "passwordsalt": "**_REMOVED SENSITIVE VALUE**_",
        "secret": "**_REMOVED SENSITIVE VALUE**_",
        "trusted_domains": [
            "**_REMOVED SENSITIVE VALUE**_"
        ],
        "datadirectory": "\/srv\/nextcloud\/data",
        "version": "9.1.0.16",
[...]

PR #740, opened 2016-09-13
add fallback local names to fullcalendar

georgehrke:
partial fix for #730


js/config/config.js
```diff
@@ -31,6 +31,9 @@ app.config(['$provide', '$httpProvider',
 			allowXName: true
 		};
 
+		const locale = moment.locale();
+		moment.locale([locale, 'en']);
```
georgehrke: this should ideally go into the server and not the calendar app

tcitworld: Getting default server language ? :+1:

PR #786, opened 2017-04-03
Fix phpunit version to be used

DeepDiver1975:


PVince81:
already 3 months old, so old that the Travis job log doesn't load for me.

try again differently

PVince81:
and all tests failed

PR #797, opened 2017-05-22
IT l10n: fix translation

soujak:
Expansion of `time` variable begins with the weekday. Old translation was therefore grammatically incorrect.


DeepDiver1975:
Please change translations via tranisfex: https://www.transifex.com/owncloud-org/owncloud/calendar/

soujak:
I would do so, but the translation project seems inactive. My request to join the translation team has been pending since more than three weeks.

Any alternative suggestion?

PR #893, opened 2018-02-28
Bump max version to 10.1 for core master changes

phil-davis:
and adjust Travis matrix.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,7 +24,7 @@ branches:
 
 env:
   global:
-    - CORE_BRANCH=master
+    - CORE_BRANCH=stable10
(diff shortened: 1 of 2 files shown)


phil-davis:
but maybe you do not want to bump max version to 10.1 just yet in ``info.xml``
we could just do the switch to default testing against ``core`` ``stable10`` and not add the extra Travis matrix entries for testing against ``core`` ``master``

DeepDiver1975 (approved):

DeepDiver1975:
> but maybe you do not want to bump max version to 10.1 just yet in info.xml

no - this is fine

PR #895, opened 2018-02-28
Escape calendar sharing drop down.

DeepDiver1975:
After spending the evening with @Julian1998 to understand angular, typeahead and all related fun: the solution was quite easy ;-)


Julian1998:
Works fine! But fun fact:
If you want to share with test1 and you are searching for <img.... it isn't highlighted anymore

Julian1998 (commented):
Thx

Julian1998 (approved):

PR #922, opened 2018-06-14
Updated some css and translations

Julian1998:


templates/part.eventsrepeat.php
```diff
@@ -54,13 +54,13 @@ class="pull-right pull-half"
 	</label>
 	<div class="pull-right pull-half">
```
DeepDiver1975: if I remove .pull-half it looks like this .... a bit better from my pov and consumes less or equal space

screenshot from 2018-06-14 21-01-12

@Julian1998 what do you think?

DeepDiver1975 (changes requested):
- see comment
- you wanted to remove the weekday picker in some cases - different PR?

templates/editor.sidebar.php
```diff
@@ -113,7 +113,7 @@ class="events--button button btn delete btn-half pull-left"
 						</button>
 						<ul class="dropdown-menu" uib-dropdown-menu role="menu" aria-labelledby="single-button">
 							<li role="menuitem"><a ng-click="delete()" href="#"><?php p($l->t('Delete all')); ?></a></li>
-							<li role="menuitem"><a ng-click="deleteOccurrence()" href="#"><?php p($l->t('Delete just occurrence')); ?></a></li>
+							<li role="menuitem"><a ng-click="deleteOccurrence()" href="#"><?php p($l->t('Delete this event')); ?></a></li>
```
DeepDiver1975: Maybe we need to add the date and time of the explicit event.
After opening one event you basically get the series editor - the relation to 'this event' is missing

DeepDiver1975 (changes requested):

templates/editor.sidebar.php
Julian1998: You are right! I would suggest that we "fix" the series editor. If you click on any repeating event you will  just see the start and end time of the first event... So every occurrence should have it's own start and end time... But I think that's a way bigger problem because datetimepicker initialization and fullcalendar have no possibility to work together right now 

-> new PR

templates/part.eventsrepeat.php
Julian1998: I will update this PR in some days... ;)

Julian1998:
We have to limit recurrence rule combinations:

* ByDay is not working in combination with monthly recurrence if interval is set to 1
* A daily recurrence with byday option set is useless so remove byday for daily recurrences
* It's the same for yearly recurrences -> Remove byday for yearly recurrences

templates/editor.sidebar.php
Julian1998: https://github.com/owncloud/calendar/issues/923

DeepDiver1975:
- we shall add some explanatory text to each simplified repeat option.

DeepDiver1975 (changes requested):
- [x]  more descriptive text in option dropdown
- [x]  text alignment on repeat tab
- [x]  custom tab: one line "Repeat every [int] [days|...]"
- [x]  alignment of end
- [x] disable not selected end options
- [x] no weekdays on months
- [x] end options: never, after [int] events, at [datepicker]
- [x]  no time on end date picker
- [x]  on monthly: every [1|2|3|4|last] [monday|...]
- [x]  delete button: "Delete this event (date)"

Julian1998:
- [ ] test

DeepDiver1975:
- [x] we have to respect localization of the day. It is 23. in German but 23rd in English - maybe mometjs can help?

DeepDiver1975:
see https://momentjs.com/docs/#/displaying/format/ - use format string Do

DeepDiver1975:
- [ ] left alligning the two inputs fields on the end section would be great - like google calendar does ...

DeepDiver1975 (approved):

PR #942, opened 2018-09-11
Fix wrong highlighted today marker

Julian1998:
.fc-state-highlight has been removed for some reason... so we have use a different css auto generated css class


DeepDiver1975 (approved):

Julian1998:
@nonasoumy Not sure what you mean by "those files are empty"...

PR #946, opened 2018-09-14
EXDATE and DTSTART have to have the same format. Before saving an eve…

DeepDiver1975:
…nt EXDATE values are recomputed based on the format of DTSTART

### DIFF
--- a/tests/js/unit/models/veventModelSpec.js
+++ b/tests/js/unit/models/veventModelSpec.js
@@ -2050,4 +2050,90 @@ END:VEVENT`.split("\n").join("\r\n"));
 
 		expect(called).toEqual(true);
 	});
+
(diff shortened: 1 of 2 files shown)


tests/js/unit/models/veventModelSpec.js
```diff
@@ -2050,4 +2050,90 @@ END:VEVENT`.split("\n").join("\r\n"));
 
 		expect(called).toEqual(true);
 	});
+
+	var using = function (values, func) {
```
DeepDiver1975, line 2054: @PVince81 this is some data provider code I was taking from the npm package jasmine-data-provider.
I could not get it to work properly ....
Is it worth putting this some there else? THX

PR #964, opened 2018-11-04
Fix for exdate timezone recalculations

Julian1998:
Fix for https://github.com/owncloud/enterprise/issues/2918

Test still failing

How to test:
1. Create a new event
2. create a recurrence
3. toggle allday
4. Exdates should still appear in both directions

### DIFF
--- a/js/app/models/veventModel.js
+++ b/js/app/models/veventModel.js
@@ -293,7 +293,10 @@ app.factory('VEvent', function(TimezoneService, FcEvent, SimpleEvent, ICalFactor
 
 			const exdateProp = new ICAL.Property('exdate', vevent);
 			exdateProp.setValues(exdates);
-			if (angular.isDefined(dtstart.timezone)) {
+			if (angular.isDefined(dtstart.zone) && dtstart.zone.tzid !== 'floating' &&
+				dtstart.zone.tzid !== 'UTC') {
+				exdateProp.setParameter('tzid', dtstart.zone.tzid);
+			} else if(angular.isDefined(dtstart.timezone)) {
 				exdateProp.setParameter('tzid', dtstart.timezone);
 			}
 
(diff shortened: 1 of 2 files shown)


Julian1998:
@DeepDiver1975 Pls review!

phil-davis:
This will need a rebase to get a current (and hopefully passing) CI result.

phil-davis:
@ownclouders rebase

DeepDiver1975:
As far as my debugging goes for today: the parser is having issues in properly setting the timezone .... :man_shrugging:

karakayasemi:
@ownclouders rebase

karakayasemi:
The parser cannot parse timezone in `dtstart` property. There are several open issues about that in `ical.js` repo like https://github.com/mozilla-comm/ical.js/issues/257 . That's why js tests are failing. I will apply one of the suggested ways to hack this problem.

DeepDiver1975:
@karakayasemi what did you change to make it work now? Awesome! THX

karakayasemi:
@DeepDiver1975 @Julian1998  I am not really familiar with ical protocol, your review helps a lot in here. As far as I understand, here are my observations about the problem:

- When deleting one of the event occurrences, the event's `dtstart` property has both `timezone` and `zone.tzid` properties with the same value. However, when updating the same event status, the parser was not setting `timezone` property for `dtstart`, only it was setting  `zone.tzid`. That's why @Julian1998's solution works in the production environment for https://github.com/owncloud/enterprise/issues/2918 and https://github.com/owncloud/enterprise/issues/3248 issues. 

- However for test cases, when `dtstart` has `tzid` property but `exdate` has not `tzid` property, ical.js cannot parse `zone.tzid` for `dtstart`. This situation was blowing up one of the tests after @Julian1998 commit. Actually, I could not reproduce this scenario in production, but I added an `elseif` to use `timezone` for this case.

Julian1998:
@karakayasemi Test works fine now! Afaik timezone is just a variable set by our own and not belongig to the ical protocol. But as long as it works hehe :+1:

DeepDiver1975 (approved):

PR #978, opened 2019-02-12
Fix TRIGGER:-P - just like we fixed TRIGGER:P

DeepDiver1975:
Fixes invalid ics with TRIGGER:-P ... similar to TRIGGER:P


DeepDiver1975:
@georgehrke FYI

PR #1077, opened 2020-02-08
Update starlark code

phil-davis:
https://github.com/owncloud/QA/issues/640

Note: this repo is in the process of converting from Travis to Drone. This adds the full "standard" set of drone starlark code.

I did not switch to `.drone.star` because it might be easier for the rest of the conversion from Travis to stay having `.drone.yml`

### DIFF
--- a/.drone.starlark
+++ b/.drone.starlark
@@ -19,7 +19,7 @@ config = {
 		},
 	},
 
-	'javascript': True,
+	'javascript': False,
 
 	'phpunit': {
 		'allDatabases' : {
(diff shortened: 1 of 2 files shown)


.drone.starlark
```diff
@@ -19,7 +19,7 @@ config = {
 		},
 	},
 
-	'javascript': True,
+	'javascript': False,
```
phil-davis, line 22: Note: before this PR there was no starlark code that executed JS tests, so having this set to True had no effect.

I have set it to false for now, because the JS tests have not yet been looked at to run on drone. That is a separate job.

individual-it (approved):

Issue #1079, opened 2020-02-10
Standardize of Makefile has lost the detail of make source and appstore


phil-davis:
PR #1055 modified `Makefile` to make it "standard" similar to other apps.

But actually there was a lot of code related to the `source:` and `appstore:` targets that selected specific files to be put in the output. That needs to be put back the way it was.

phil-davis:
@DeepDiver1975 I can just put it back the way it was.

But for my education, what is the purpose of having the `source:` and `appstore:` "make targets"?

(maybe nobody cares about the `source:` target? or?)

phil-davis:
Fixed by PR #1080

PR #1055, opened 2020-01-08
makeFile modified

haribhandari07:
## Description
This PR adjusts `Makefile` for the necessary `phpunit` and `javascript` dependencies. 
This PR is the first step of moving `calendar` app's `ci` from `travis` to `drone`

### Related Issues
https://github.com/owncloud/calendar/issues/935

PR #1080, opened 2020-02-11
Recover makefile appstore

phil-davis:
Fixes issue #1079 

- Revert the `Makefile` changes that were made in PR #1055 - that gets us back to the original `dist` `source` `appstore` etc targets.
- Add/adjust test targets in the format similar to how it is done in other apps, so that code-style, PHP unit tests and JS tests can run "standard" in drone. (without touching any of the existing targets related to dist/source/appstore...)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,36 +1,74 @@
-SHELL := /bin/bash
+# This file is licensed under the Affero General Public License version 3 or
+# later. See the COPYING file.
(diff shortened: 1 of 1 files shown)


haribhandari07 (commented):
LGTM

DeepDiver1975 (approved):

Issue #935, opened 2018-09-05
Drone setup


davitol:
Move from Travis to Drone

phil-davis:
Added to sprint backlog. Maybe this can go in the next sprint - should be easy enough to do now that we have "standard" drone starlark etc.

phil-davis:
@HariBhandari07 I switched https://drone.owncloud.com/owncloud/calendar/settings to `.drone.star`
You can make a PR to rename `.drone.starlark` to `.drone.star` and delete `.drone.yml`

PR #1082, opened 2020-02-17
[Tests-only] run js tests on drone

haribhandari07:
### Description
Run js test on drone

Issue #935

### DIFF
--- a/.drone.starlark
+++ b/.drone.starlark
@@ -19,7 +19,15 @@ config = {
 		},
 	},
 
-	'javascript': True,
+	'javascript': {
+		'extraCommandsBeforeTestRun': [
+			'apt update -y',
+			'apt-get install xvfb libxrender1 libxtst6 libxi6 -y',
+			'/usr/bin/Xvfb :99 &',
+			'export DISPLAY=:99.0',
+			'apt-get install firefox -y',
+		],
+	},
 
 	'phpunit': {
 		'allDatabases' : {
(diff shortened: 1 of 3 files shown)


phil-davis:
Note: Travis fails because of https://github.com/Graffino/browser-update dependency tag problems.

See my comments in https://github.com/Graffino/browser-update/pull/1

.drone.starlark
```diff
@@ -477,6 +477,11 @@ def javascript():
 				'pull': 'always',
 				'environment': params['extraEnvironment'],
 				'commands': params['extraCommandsBeforeTestRun'] + [
+					'apt update -y',
+					'apt-get install xvfb libxrender1 libxtst6 libxi6 -y',
+					'/usr/bin/Xvfb :99 &',
+					'export DISPLAY=:99.0',
+					'apt-get install firefox -y',
```
phil-davis: These commands should be able to go in the `javascript` section of `config` up the top, see an example in `files_antivirus` `.drone.star` (where `extraCommandsBeforeTestRun` is used for `phpunit`)

Then we do not have different starlark code in this repo.

phil-davis (commented):
And we will have to wait to see what happens for the https://github.com/Graffino/browser-update upstream problem before we can get a Travis run to make sure Travis still passes.

zenopopovici:
@phil-davis  Change has been reverted for https://github.com/Graffino/browser-update. Sorry for the trouble.

phil-davis (approved):

phil-davis:
@HariBhandari07 now we should switch from `.drone.yml` to `.drone.star`

When you are ready to make a PR for that, let me know and I will change the drone server settings to expect `.drone.star`

PR #1088, opened 2020-02-25
All time zone information is taken from the timezone service which gr…

DeepDiver1975:
…abs information from the zones.json file

In case a timezone is unknown it is taken from the event itself.

alternative to #1087


DeepDiver1975:
I'm putting this on hold ..... the explicit case which we a re trying to fix is most likely an upstream issue in https://github.com/mozilla-comm/ical.js/pull/421

PR #1150, opened 2020-12-01
[Tests-Only][CI] fix the starlark code for reduced ci after merge and sonarcloud

dpakach:
fix the starlark code for reduced ci after merge and sonarcloud

issue https://github.com/owncloud/QA/issues/649

### DIFF
--- a/.drone.star
+++ b/.drone.star
@@ -51,33 +51,58 @@ config = {
 def main(ctx):
 	before = beforePipelines()
 
-	stages = stagePipelines()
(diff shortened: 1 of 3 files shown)


phil-davis:
I added a commit to adjust the location of the code to be analysed by SonarCloud.

The javascript tests do not have coverage turned on, so there will not be coverage for those. But I included them in `sonar.inclusions` because then the Sonar analysis will at least process them and report any code issues that it finds.

phil-davis (approved):

PR #1162, opened 2021-01-25
[docs-only] Remove unused badges from README

phil-davis:
PR #1161 added to SonarCloud badges to README.There is still an old badge to also remove.


swoichha (approved):
LGTM

PR #1212, opened 2021-10-21
[tests-only] Add compatibility with Guzzle 7

JammingBen:
https://github.com/owncloud/core/pull/39368 bumps Guzzle to v7 in core. These changes are necessary to make it work with this app.


phil-davis:
Needs guzzle7 merged to core so that tests will work.

phil-davis (approved):

phil-davis:
Note: only changes to unit tests were required. So the actual existing code of the app may work fine with core 10.11 that has Guzzle7. Needs some manual testing to see if a release is needed.

PR #1254, opened 2022-05-19
[tests-only][full-ci]Tidy up the phpunit xml file for calendar app

SagarGi:
This PR moves unit tests into tests/unit folder. And also tidy up the `phpunit.xml` to make standard format for all oc-apps as much as possible.

- Part of https://github.com/owncloud/impersonate/issues/198


phil-davis (approved):
Note: for this app, the unit tests were already in `tests/unit` so they did not need to be moved.

PR #1258, opened 2022-07-01
Bump owncloud min-version to 10.11 for Guzzle7

phil-davis:


phil-davis:
This happened in PR #1273

